PR #21, opened 2023-05-08
Create Blank Ruby on Rails Service

### LINKED ISSUE #4: Create a new blank rails service for the registration system
- [x] Use the Dockerfile from the current WcaOnRails Service but strip it down to the bare minimum

QUESTIONS:
1. Will we have separate repos for the frontend and backend? Or how is the React side of things being included?
2. Friend of mine (experienced React dev) suggested we use ViteJS to set up a Rails/React project - might be worth looking into?

FinnIckler:
I created a blank Rails service using the `rails new wca-registration --api` cli.

There might be some not needed gems in there, but we will find out later.

You can run it with `docker compose up`

Closes #4


dunkOnIT:
LGTM

dunkOnIT (approved):

PR #25, opened 2023-05-09
Create Infrastructure and deployment for Service using Terraform

### LINKED ISSUE #3: Create a devops pipeline to deploy the new service to AWS on backend.worldcubeassocation.org/registration or registration.worldcubeassociation.org
- [ ] Decide on URL

FinnIckler:
This creates a bunch of infrastructure to deploy the registration service including
- A VPC to house all the infrastructure
- An ECS cluster to manage container placement
- A EC2 instance acting as the host for the container
- A Load Balancer to route the traffic to the cluster
- A Codepipeline to manage deployment on image push

I basically stole it all from [Jonatan's deployment of WCA Live](https://github.com/thewca/wca-live/tree/main/infra) with a few things removed that I didn't need. 

I have ran this through and you can now access the service on https://registration.worldcubeassociation.org/ (you just get an error but that is intentional)

Included is a deployment action also taken from wca-live which builds the image and pushes it to the Amazon Elastic Container Registry which in turn triggers the Code Pipeline

Closes #3

### DIFF
--- a/infra/app.tf
+++ b/infra/app.tf
@@ -1,0 +1,411 @@
+resource "aws_security_group" "cluster" {
+  name        = "${var.name_prefix}-cluster"
+  description = "Production ECS cluster"
+  vpc_id      = aws_vpc.this.id
+
+  tags = {
+    Name = "${var.name_prefix}-cluster"
+  }
+}
+
+# Note: we use the standalone SG rules (rather than inline), because
+# cluster_cluster_ingress references the SG itself
+
+resource "aws_security_group_rule" "cluster_lb_ingress" {
+  type                     = "ingress"
+  security_group_id        = aws_security_group.cluster.id
+  from_port                = 0
+  to_port                  = 0
+  protocol                 = "-1"
+  source_security_group_id = aws_security_group.lb.id
+  description              = "Load balancer ingress"
+}
+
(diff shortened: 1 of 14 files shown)


infra/app.tf
```diff
@@ -0,0 +1,411 @@
...
+  network_mode             = "awsvpc"
+  requires_compatibilities = ["EC2"]
+
+  # We configure the roles to allow `aws ecs execute-command` into a task,
+  # as in https://aws.amazon.com/blogs/containers/new-using-amazon-ecs-exec-access-your-containers-fargate-ec2
+  execution_role_arn = aws_iam_role.task_execution_role.arn
+  task_role_arn      = aws_iam_role.task_role.arn
+
+  cpu = "1024"
+  memory = "1024"
```
jonatanklosko, line 154: I think if the instance itself has `2048` then it may not fit two containers, because the instance itself also uses a small bit of memory. Not sure what's the optimal value (800-900?), but just pointing out that blue/green would likely spawn a separate instance.

jonatanklosko (approved):

infra/app.tf
FinnIckler: makes sense, these values are chosen at random now anyway. I created an issue here #26 to revisit once we have something to load test.

PR #28, opened 2023-05-09
Add healthcheck route

### LINKED ISSUE #27: Create Healthcheck route for scaling and deploying

FinnIckler:
Added a /healthcheck path which is useful full deployment and scaling.  
I also used this change to test the pipeline and everything worked (after tweaking the memory values down a bit). Seems to take about 10 minutes, but half of that is just waiting in case we want to abort the deployment if there are some issues. 

Closes #27

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,4 +5,4 @@ under development. For more info about the WCA visit the main repo [here](https:
 
 ## How to run
 
-Run  `docker compose up`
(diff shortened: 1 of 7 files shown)


FinnIckler:
Merging because Duncan is out of office this morning. This is all just part of the PoC anyway.

PR #30, opened 2023-05-10
Add dummy routes for /register and /registrations

FinnIckler:
Creates Two Dummy routes:
- /register:  does very basic validation for WCA id and Competition id and inserts them into the database
- /registration: load registrations from the database by competition_id

The database used is a DynamoDB Table with the schema
id : uuid
competition_id: string
competitor_id: string
registration_data: json object

The keys for the Table are competition_id (hash key) and competitor_id (range key). 

Locally, the database is simulated using the official dynamodb-local image and I added the dynamodb-admin image to look into the DB if needed.

I also created http://test.registration.worldcubeassociation.org.s3-website-us-west-2.amazonaws.com/ for some simple frontend testing

### DIFF
--- a/app/controllers/healthcheck_controller.rb
+++ b/app/controllers/healthcheck_controller.rb
@@ -1,0 +1,5 @@
+class HealthcheckController < ApplicationController
+  def index
+    render json: { status: "ok" }, status: :ok
+  end
+end
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,9 @@
+node_modules
+.git
+.github
+tmp/*
+log/*
+infra
+storage/*
+docker-compose.dev.yml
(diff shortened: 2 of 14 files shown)


app/controllers/healthcheck_controller.rb
dunkOnIT, line 1: What is healthcheck?

dunkOnIT:
JGTM just minor nits (won't be able to pick up on if anything is wrong with the config files tho :D )

app/controllers/healthcheck_controller.rb
FinnIckler: It's a route that just returns 200 ok for deployment.
In the future we should probably include checks to know if everything works correctly

dunkOnIT (approved):

PR #43, opened 2023-05-16
Create a React Frontend and deployment for it

### LINKED ISSUE #23: Rewrite the Register Form in React
We might be able to contribute this back to the main repo if just for the reason to half the amount of calls for registration

FinnIckler:
I created a simple React frontend using my previous testing HTML file.
- I use esbuild to bundle the js and css files
- I use (s)css.modules to create styles and reference classnames directly in react

For deployment I created a cloudfront distribution which serves the assets.registrations.worldcubeassociation.org Bucket where the bundles are deployed two. After deployment the cache is invalidated. 

closes #23


Frontend/src/api/registration/delete/delete_registration.js
```diff
@@ -0,0 +1,20 @@
+export default async function delete_registration(competitor_id, competition_id){
+    const formData = new FormData();
+    formData.append('competitor_id', competitor_id);
+    formData.append('competition_id', competition_id);
+
+    try {
```
coder13, line 6: We should extrapolate the try catch into it's own function. We can a `MyFetch` as a function that fetches data from the backend and does the try catch and error formatting.

coder13: I also don't believe we need a separate directory for each file. These files are already scoped to the registration feature.

coder13:
Strong recommendation: Separate components from pages, have `pages/` and `components/` be directories in the same directory.

Frontend/src/watch.js
```diff
@@ -0,0 +1,32 @@
+const esbuild = require('esbuild');
```
coder13, line 1: I've never used esbuild but I love how simple it seems.

Frontend/package.json
```diff
@@ -0,0 +1,25 @@
...
+    "url": "git+https://github.com/thewca/wca-registration.git"
+  },
+  "author": "WCA Software Team",
+  "license": "GPL-3.0-or-later",
+  "bugs": {
+    "url": "https://github.com/thewca/wca-registration/issues"
+  },
+  "homepage": "https://github.com/thewca/wca-registration/Frontend/README.md",
+  "dependencies": {
+    "esbuild": "^0.17.19",
```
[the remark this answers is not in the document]
kr-matthews: Possibly also `eslint-plugin-react` and `eslint`? I'm not very familiar with dependencies but I use them in my projects.
Wouldn't these be dev dependencies though?

Frontend/src/api/registration/delete/delete_registration.js
```diff
@@ -0,0 +1,20 @@
+export default async function delete_registration(competitor_id, competition_id){
```
kr-matthews, line 1: I think typically the convention in js (and react) is to use camel case. It looks odd in the jsx file to be calling the function `delete_registration` when everything else is camel case. Or is that intentional to emphasize something?

FinnIckler: not intentional, just me derping when jumping between ruby and js

Frontend/package.json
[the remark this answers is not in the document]
FinnIckler: checking it out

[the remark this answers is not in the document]
FinnIckler: Should there even still be dev dependencies if all node does is build the bundle? I guess it would slightly decrease the npm install in deploy.yaml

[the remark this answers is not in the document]
FinnIckler: Addes eslint and react plugins

Frontend/src/register/components/registration_list.jsx
```diff
@@ -0,0 +1,114 @@
+import React, { useState } from 'react'
+import deleteRegistration from '../../api/registration/delete/delete_registration'
+import getRegistrations from '../../api/registration/get/get_registrations'
+import updateRegistration from '../../api/registration/patch/update_registration'
+import styles from './list.module.scss'
+
+function StatusDropdown({ status, setStatus }) {
```
coder13, line 7: Can we try to break up each component into it's own file?

Frontend/src/register/pages/index.jsx
```diff
@@ -0,0 +1,13 @@
+import React from 'react'
+import styles from './index.module.scss'
+import RegistrationList from '../components/registration_list'
+import RegistrationPanel from '../components/registration_panel'
+
+export default function App() {
+  return (
+    <div className={styles.container}>
+      <RegistrationPanel> </RegistrationPanel>
```
coder13, line 9: Nit: you could just use a self closing tag `<RegistrationPanel />`

Frontend/package.json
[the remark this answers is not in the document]
kr-matthews: I don't see the `eslint-plugin-react` or `eslint-plugin-react-hooks` plugins?

Frontend/src/register/components/registration_list.jsx
kr-matthews: It's also standard to name the file after the default exported function, so `RegistrationList.jsx` in this case (as opposed to `registration_list.jsx`).

Frontend/src/register/components/registration_list.jsx
```diff
@@ -0,0 +1,114 @@
...
+  const [competitionID, setCompetitionID] = useState('HessenOpen2023')
+  const [registrationList, setRegistrationList] = useState([])
+  return (
+    <div className={styles.list}>
+      <button
+        onClick={async (_) =>
+          setRegistrationList(await getRegistrations(competitionID))
+        }
+      >
+        {' '}
```
kr-matthews, line 74: You've got a lot of `{' '}` throughout all the components, I assume that was the linting and they aren't actually needed.

Frontend/src/register/pages/index.jsx
kr-matthews: One of the react plugins should be able to enforce this.

Frontend/src/register/components/RegistrationList.jsx
```diff
@@ -0,0 +1,101 @@
...
+      <td>
+        <StatusDropdown status={status} setStatus={setStatus} />
+      </td>
+      <td>
+        <button
+          onClick={(_) => {
+            updateRegistration(competitorId, competitionID, status)
+          }}
+        >
+          {' '}
```
coder13, line 31: Nite: you shouldn't need this space here

PR #55, opened 2023-05-31
Switch Frontend -> Backend Communication to JSON

### LINKED ISSUE #52: Add support for json submission to //register in POC

FinnIckler:
This switches all the Backend Communication to use json instead of formdata. I actually didn't have to change anything in the backend because the ActionDispatch Module handling json was already included. 

This PR also changes all the functions that handle API interactions to typescript. This is because Typescript types for the API contracts will be auto generated out of the openapi spec to always keep parity between the frontend and backend.

closes #52


Frontend/src/api/helper/backend_fetch.ts
```diff
@@ -0,0 +1,34 @@
+type Method = 'POST' | 'GET' | 'PATCH' | 'DELETE'
+
+type Body =
```
coder13, line 3: I don't think this will scale. Rather backendFetch should take any in as Body and each function that uses backendFetch should define it's own body props.

coder13: Or TS also likes using `unknown` in these situations as opposed to `any`

Frontend/src/api/types.d.ts
```diff
@@ -0,0 +1,41 @@
+// These will be generated by the swagger file generated with Rswag
+type WCAEvent =
```
coder13, line 2: Can we try to use `@WCA/Helpers` for this project? It seems like a no brainer.

coder13: To clarify, `@WCA/Helpers` already has an `EventId` type it exports.

Frontend/src/api/helper/backend_fetch.ts
FinnIckler: Yes, this is not meant to scale. All these types will be generated automatically from our API Spec

Frontend/src/api/types.d.ts
FinnIckler: ah cool, will use

coder13 (approved):

PR #64, opened 2023-06-05
Add Fomantic UI

FinnIckler:
This PR adds fomantic to the Frontend and imports the Header and Footer Components styled after the Main Website


Frontend/src/index.jsx
```diff
@@ -1,35 +1,47 @@
...
-      <div>
-        <Link to="/register"> Register </Link>
-        <Link to="/registrations"> Registrations</Link>
-      </div>
+      <>
+        <PageHeader />
+        <PageFooter />
+      </>
     ),
   },
```
kr-matthews, line 22: Can avoid repeating the header and footer with something like
```
{
  path: '/',
  element: <><PageHeader /><Outlet /><PageFooter /></>,
  children: [
    { path: '/register', element: <main><Register /></main>},
    { path: '/registrations', element: <main><Registrations /></main> },
  ],
```

FinnIckler: done

PR #72, opened 2023-06-14
Frontend: Create basic Layout and routing

FinnIckler:
This PR creates the basic routing that the Frontend should have and integrates the RegistationPanel and RegistrationList Components.

Routes:
- A dummy page at /
- A dummy page at /:competition_id
- A pretty close to done Competitor List at  /:competition_id/registrations/
- A simple Registration Edit page at /:competition_id/registrations/edit
- A simple Single Registration Edit page at /:competition_id/:userid/edit
- A simple Register page at  /:competition_id/register

All functionality should be available at https://test-registration.worldcubeassociation.org/ after I deployed to staging


FinnIckler:
This version depends on a future version of wca-components, will fix in the morning after I make a PR there

Frontend/src/pages/registrations/components/RegistrationList.jsx
```diff
@@ -1,101 +1,98 @@
...
-          }}
-        >
-          Delete
-        </button>
-      </td>
-    </tr>
-  )
-}
 
 export default function RegistrationList() {
```
kr-matthews, line 9: I think some/most of the state in here, like header and footer, can just be computed from other state instead of having its own state. In fact, there could probably be just two states, for `competitionInfo` and `registrations` with everything else derived from these. This seems safer, because theoretically all these current states could get out-of-sync as they are, whereas if they're calculated from the single source(s) of truth then they're guaranteed to be in-sync. And separate constants could be calculated separately, instead of all crammed into a couple of large use effects (which could just receive the data and call setters).
Generally, only using state for something that needs to be stateful is a good rule of thumb, I think.

Frontend/src/pages/registration_administration/components/StatusDropdown.jsx
```diff
@@ -0,0 +1,12 @@
+import React from 'react'
+
+export default function StatusDropdown({ status, setStatus }) {
+  const options = ['waiting', 'accepted', 'deleted']
```
kr-matthews, line 4: It's nice to pull constants like this out of the component, otherwise a fresh array & reference is created each render.

Frontend/src/pages/registration_administration/components/RegistrationAdministrationList.jsx
```diff
@@ -0,0 +1,101 @@
...
+          }}
+        >
+          Delete
+        </button>
+      </td>
+    </tr>
+  )
+}
+
+export default function RegistrationAdministrationList() {
```
kr-matthews, line 51: Keeping the default function at the top is good; function dependencies pointing downward in a file is nice.

Frontend/src/pages/registration_administration/components/RegistrationAdministrationList.jsx
```diff
@@ -0,0 +1,101 @@
...
+      <td>
+        <StatusDropdown status={status} setStatus={setStatus} />
+      </td>
+      <td>
+        <button
+          onClick={(_) => {
+            updateRegistration(competitorId, competitionID, status)
+          }}
+        >
+          {' '}
```
kr-matthews, line 31: Extra space?

Frontend/src/pages/registrations/components/RegistrationList.jsx
FinnIckler: Done!

Frontend/src/index.jsx
```diff
@@ -1,41 +1,61 @@
...
 import PageHeader from './ui/Header'
+import PageSidebar from './ui/Sidebar'
+import HomePage from './pages/home'
 
 const router = createBrowserRouter([
   {
     path: '/',
     element: (
       <>
         <PageHeader />
```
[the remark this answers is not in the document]
FinnIckler: done!

Frontend/src/pages/register/components/RegistrationPanel.jsx
```diff
@@ -24,24 +30,15 @@ export default function RegistrationPanel() {
...
-      </label>
       <EventSelector
         handleEventSelection={handleEventSelection}
-        events={EVENTS}
+        events={heldEvents}
         initialSelected={[]}
         size="2x"
       />
       <button
         onClick={(_) =>
```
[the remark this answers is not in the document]
FinnIckler: done!

Frontend/src/pages/registration_administration/components/RegistrationAdministrationList.jsx
```diff
@@ -1,101 +1,127 @@
...
+      <td>{name}</td>
+      <td>{country}</td>
+      <td>{new Date(registeredOn).toLocaleDateString()}</td>
+      <td>{numberOfEvents}</td>
       <td>
         <StatusDropdown status={status} setStatus={setStatus} />
       </td>
       <td>
         <button
+          // TODO Update the list automatically
```
kr-matthews, line 117: Will look at later.

Frontend/src/pages/registrations/components/RegistrationList.jsx
```diff
@@ -8,60 +8,90 @@ import styles from './list.module.scss'
...
-      const rows = registrations.map(async (registration) => {
+      const regList = []
+      for (const registration of registrations) {
+        registration.user = (
+          await getCompetitorInfo(registration.competitor_id)
+        ).user
+        regList.push(registration)
+      }
+      setRegistrations(regList)
+      setLoading(false)
```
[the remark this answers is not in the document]
FinnIckler: I was trying to just use Suspense but that does not seem to work

Frontend/src/pages/registration_administration/components/RegistrationAdministrationList.jsx
```diff
@@ -1,101 +1,127 @@
...
+    </div>
+  )
+}
+
+function RegistrationAdministrationTable({ registrations }) {
+  return (
+    <table className="table">
+      <thead>
+        <tr>
+          <th> WCA ID</th>
```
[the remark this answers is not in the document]
FinnIckler: Nope, no idea how that happened

Frontend/src/pages/registration_administration/components/RegistrationAdministrationList.jsx
```diff
@@ -1,101 +1,127 @@
...
+      for (const registration of registrations) {
+        registration.user = (
+          await getCompetitorInfo(registration.competitor_id)
+        ).user
+        regList.push(registration)
+      }
+      setRegistrations(regList)
+    })
+  }, [competition_id])
+  const partitionRegistrations = (registrations) => {
```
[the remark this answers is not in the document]
FinnIckler: done

Frontend/src/pages/registrations/components/RegistrationList.jsx
```diff
@@ -8,60 +8,90 @@ import styles from './list.module.scss'
...
+      `${countrySet.size} Countries`,
+      ...eventCounts.values(),
+      total,
+      '',
+    ]
+  }, [registrations, heldEvents])
+  const registrationList = useMemo(
+    () =>
+      registrations.map((registration) => {
         let profile_link = null
```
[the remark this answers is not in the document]
FinnIckler: done

Frontend/src/pages/registrations/components/RegistrationList.jsx
```diff
@@ -8,60 +8,90 @@ import styles from './list.module.scss'
...
+  }, [competition_id])
+
+  const header = [
+    { text: 'Name' },
+    { text: 'Citizen of' },
+    ...heldEvents.map((event_id) => ({ text: event_id, cubingIcon: true })),
+    { text: 'Total' },
+  ]
+  const footer = useMemo(() => {
+    let newcomers = 0
```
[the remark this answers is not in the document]
FinnIckler: I put it in one reduce to make it const, but still only iterate over the array once.

Frontend/src/pages/registration_administration/components/RegistrationAdministrationList.jsx
```diff
@@ -0,0 +1,134 @@
...
+                country={registration.user.country.name}
+                registeredOn={registration.registered_on}
+                numberOfEvents={registration.event_ids.length}
+                initialStatus={registration.registration_status}
+              />
+            )
+          })
+        ) : (
+          <tr>
+            <td colSpan={6}>No matching records found</td>
```
kr-matthews, line 94: Won't this show up while the data is still loading?
I guess that's another advantage to having the initial data set to `undefined` rather than `[]`: so you can easily distinguish in sub-components whether it's still loading or has finished with no results, without having to also pass `isLoading` prop everywhere.

FinnIckler: I have the isLoading in the parent return, so it shouldn't

FinnIckler:
@thewca-bot deploy staging

FinnIckler:
All Components added, please review

Frontend/src/pages/registration_administration/components/RegistrationAdministrationList.jsx
```diff
@@ -0,0 +1,261 @@
...
+    },
+    { waiting: [], accepted: [], deleted: [] }
+  )
+}
+
+export default function RegistrationAdministrationList() {
+  const { competition_id } = useParams()
+  const [registrations, setRegistrations] = useState([])
+  const [isLoading, setIsLoading] = useState(true)
+  const [selected, setSelected] = useState({
```
kr-matthews, line 37: Might be worth using a reducer here (via `useReducer`) so that the logic of updating the arrays can be separated into the reducer helper function, and the UI below can look simpler, something like
```
          add={(attendee) => dispatchSelected({ action: 'add-waiting', attendee })
```
or
```
          add={(attendee) => dispatchSelected({ action: 'toggle-waiting', attendee })
```
rather than
```
          add={(attendee) =>
            setSelected({
              waiting: [...selected.waiting, attendee],
              accepted: selected.accepted,
              deleted: selected.deleted,
            })
```

Frontend/src/pages/registration_administration/components/RegistrationAdministrationList.jsx
```diff
@@ -0,0 +1,261 @@
...
+  )
+}
+
+function RegistrationAdministrationTable({
+  registrations,
+  add,
+  remove,
+  competition_id,
+}) {
+  const [checkedBoxes, setCheckedBoxes] = useState([])
```
kr-matthews, line 154: There's already state in the parent for which rows are selected right? It would be good to use that as the single source of truth, unless trying to set that up would be blocked by something?

Frontend/src/pages/registration_edit/components/registration_editor.jsx
```diff
@@ -0,0 +1,122 @@
...
+import { Button, Checkbox, TextArea } from 'semantic-ui-react'
+import getCompetitionInfo from '../../../api/competition/get/get_competition_info'
+import { getSingleRegistration } from '../../../api/registration/get/get_registrations'
+import { updateRegistration } from '../../../api/registration/patch/update_registration'
+import getCompetitorInfo from '../../../api/user/get/get_user_info'
+import { setMessage } from '../../../ui/events/messages'
+import LoadingMessage from '../../../ui/loadingMessage'
+import styles from './editor.module.scss'
+
+export default function RegistrationEditor({ user_id, competition_id }) {
```
kr-matthews, line 12: File name should match the component function name.

Frontend/src/pages/registration_edit/components/registration_editor.jsx
```diff
@@ -0,0 +1,122 @@
...
+
+  useEffect(() => {
+    getCompetitionInfo(competition_id).then((competitionInfo) => {
+      setHeldEvents(competitionInfo.event_ids)
+    })
+  }, [competition_id])
+
+  return (
+    <div className={styles.editor}>
+      {!registration.registration_status ||
```
kr-matthews, line 43: It's a bit inconsistent that here we check whether something is loading by seeing if the data exists yet, but other components we have an explicit `isLoading` state. Personally I prefer the former (less state), but either is fine. Consistency would be nice.
There's also the potential edge-case of a competition having 0 events (before it's been setup). Seems unlikely someone would need this page in that case, but technically the loading UI would be displayed forever.

Frontend/src/pages/registration_edit/components/registration_editor.jsx
```diff
@@ -0,0 +1,122 @@
...
+              setSelectedEvents(events)
+            }}
+            initialSelected={registration.event_ids}
+            events={heldEvents}
+            size="2x"
+          />
+          <h3> Comment </h3>
+          <TextArea
+            value={comment}
+            onChange={(_, data) => {
```
kr-matthews, line 61: Some onChanges here have `_`, some have `e`.

Frontend/src/pages/registrations/components/RegistrationList.jsx
```diff
@@ -1,101 +1,112 @@
...
-  eventIDs,
-  serverStatus,
-  competitionID,
-  setRegistrationList,
-  registrationList,
-}) {
-  const [status, setStatus] = useState(serverStatus)
+export default function RegistrationList() {
+  const { competition_id } = useParams()
+  const [isLoading, setIsLoading] = useState(true)
```
kr-matthews, line 11: There's potential for confusion later since there are 2 effects fetching data, but this `isLoading` refers to only one of them.

Frontend/src/pages/registrations/components/RegistrationList.jsx
```diff
@@ -1,101 +1,112 @@
...
+  const [registrations, setRegistrations] = useState([])
+  const [heldEvents, setHeldEvents] = useState([])
```
kr-matthews, line 13: This `heldEvents` state plus corresponding effect to fetch happens in a lot of components. It might be worth creating a custom hook and calling that instead:
```
const { isLoading, heldEvents } = useHeldEvents(competition_id)
```
and the hook:
```
function useHeldEvents(competition_id) {
  const [isLoading, setIsLoading] = useState(true);
  const [heldEvents, setHeldEvents] = useState(null); // or [] instead of null, as you have it now

  useEffect(() => {
    getCompetitionInfo(competition_id).then((competitionInfo) => {
      setHeldEvents(competitionInfo.event_ids);
      setIsLoading(false);
    })
  }, [competition_id])

  return { isLoading, heldEvents };
}
```

Similarly for other things which are fetched more than 2 or 3 times.

kr-matthews (commented):
Could we keep PRs smaller if possible? This took way too long to review.

Also note that I'm just looking at the code, I don't have an environment set up to run it.

FinnIckler:
Sorry for the length. Feature creeped myself adding components. Future PRs will be much shorter!

Frontend/src/api/registration/get/get_registrations.ts
```diff
@@ -1,5 +1,42 @@
+import { EventId } from '@wca/helpers'
 import backendFetch from '../../helper/backend_fetch'
 
-export default async function getRegistrations(competitionID: string) {
+type RegistrationStatus = 'waiting' | 'accepted' | 'deleted'
```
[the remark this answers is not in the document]
FinnIckler: Yes, but we will support it in a separate Staff Lane, where you will be able to sign up just as staff if the organizers enables it.

Frontend/src/api/registration/patch/update_registration.ts
```diff
@@ -1,14 +1,19 @@
+import { EventId } from '@wca/helpers'
 import backendFetch from '../../helper/backend_fetch'
 import { UpdateRegistrationBody } from '../../types'
 
-export default async function updateRegistration(
+export async function updateRegistration(
   competitorID: string,
   competitionID: string,
-  status: string
+  status?: string,
```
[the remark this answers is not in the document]
FinnIckler: agreed, done

Frontend/src/ui/flashMessage.jsx
```diff
@@ -0,0 +1,33 @@
...
+import Bus from './events/Bus'
+import styles from './flash.module.scss'
+
+export default function FlashMessage() {
+  const [visible, setVisible] = useState(false)
+  const [message, setMessage] = useState('')
+  const [type, setType] = useState('')
+
+  useEffect(() => {
+    Bus.addListener('flash', ({ message, type }) => {
```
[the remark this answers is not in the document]
FinnIckler: Bus is there to communicate events to the flashMessage component (and I want to use it for the Loading Message component too). Is there a better way to do this? I didn't want to give a setMessage function to every component.

Frontend/src/pages/registration_edit/components/registration_editor.jsx
FinnIckler: This page is not usable/reachable when the competition hasn't been setup, because it requires a registration to be present. I agree with the consistency argument, I will remove the isLoading states throughout in a different PR that introduces a global loading component

Frontend/src/pages/registration_edit/components/registration_editor.jsx
FinnIckler: changed to _ everywhere where it is not used

Frontend/src/pages/registrations/components/RegistrationList.jsx
FinnIckler: makes sense, I will fetch all the data in one useEffect.

Frontend/src/pages/registrations/components/RegistrationList.jsx
FinnIckler: did not know about this, done!

Frontend/src/ui/flashMessage.jsx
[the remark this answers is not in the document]
kr-matthews: I don't know much about this, but the only alternative I'm aware of would be to have a context which provides the `setMessage` function to any component which wants to access it, and provides the message to the flash component.

Frontend/src/pages/registration_edit/components/registrationEditor.jsx
```diff
@@ -1,27 +1,21 @@
...
 import { updateRegistration } from '../../../api/registration/patch/update_registration'
-import getCompetitorInfo from '../../../api/user/get/get_user_info'
 import { setMessage } from '../../../ui/events/messages'
 import LoadingMessage from '../../../ui/loadingMessage'
 import styles from './editor.module.scss'
 
 export default function RegistrationEditor({ user_id, competition_id }) {
-  const [competitorInfo, setCompetitorInfo] = useState({})
-  const [heldEvents, setHeldEvents] = useState([])
+  const { eventsLoading, heldEvents } = useHeldEvents(competition_id)
```
kr-matthews, line 12: Since the hook is returning an object with properties `isLoading` and `heldEvents`, `eventsLoading` will always be undefined. You need to either change the hook to return an array `[isLoading, heldEvents]`, or change this const to `{ isLoading: eventsLoading, heldEvenst }`.

FinnIckler:
Not sure why your other comments don't show up in the main thread, working on them now

Frontend/src/pages/register/components/RegistrationEditPanel.jsx
```diff
@@ -0,0 +1,67 @@
...
+  const [heldEvents, setHeldEvents] = useState([])
+  const [islLoading, setIsLoading] = useState(true)
+  const { competition_id } = useParams()
+  useEffect(() => {
+    getCompetitionInfo(competition_id).then((competitionInfo) => {
+      setHeldEvents(competitionInfo.event_ids)
+      setIsLoading(false)
+    })
+  }, [competition_id])
+  const handleEventSelection = (selectedEvents) => {
```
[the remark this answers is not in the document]
FinnIckler: inlined

Frontend/src/pages/register/components/RegistrationPanel.jsx
```diff
@@ -1,51 +1,63 @@
...
+import { Button, TextArea } from 'semantic-ui-react'
+import getCompetitionInfo from '../../../api/competition/get/get_competition_info'
 import submitEventRegistration from '../../../api/registration/post/submit_registration'
+import { setMessage } from '../../../ui/events/messages'
+import LoadingMessage from '../../../ui/loadingMessage'
 import styles from './panel.module.scss'
 
-const EVENTS = ['222', '333', '444', '555', '666', '777']
-
 export default function RegistrationPanel() {
```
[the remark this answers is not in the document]
FinnIckler: yeah, I noticed when I wrote it, I am kind scared that it will differ much more though in the future, so I will keep it like that for now,

Frontend/src/pages/registration_administration/components/RegistrationActions.jsx
```diff
@@ -0,0 +1,117 @@
...
+            return msg
+          }, ''),
+        'negative'
+      )
+    } else {
+      setMessage('Successfully saved registration changes', 'positive')
+    }
+    refresh()
+  }
+  return anySelected ? (
```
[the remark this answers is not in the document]
FinnIckler: changed

Frontend/src/pages/registration_administration/components/RegistrationActions.jsx
```diff
@@ -0,0 +1,117 @@
...
+        >
+          <UiIcon name="envelope" /> Email
+        </a>
+      </Button>
+      {anyApprovable ? (
+        <Button
+          positive
+          onClick={() =>
+            changeStatus(
+              [...selected.waiting, ...selected.accepted, ...selected.deleted],
```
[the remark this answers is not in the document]
FinnIckler: Because I was just following what the monolith does :D. Changed.

Frontend/src/pages/registration_administration/components/RegistrationAdministrationList.jsx
FinnIckler: TIL about useReducer. Works great!

Frontend/src/pages/registration_administration/components/RegistrationAdministrationList.jsx
FinnIckler: very true, changed

Frontend/src/pages/registration_edit/components/registrationEditor.jsx
FinnIckler: oops, changed

FinnIckler:
@thewca-bot deploy staging

Frontend/src/pages/register/components/RegistrationEditPanel.jsx
```diff
@@ -13,18 +13,16 @@ export default function RegistrationEditPanel({ registration }) {
...
-  }
   return isLoading ? (
     <div className={styles.panel}>
       <LoadingMessage />
     </div>
   ) : (
     <div className={styles.panel}>
       <EventSelector
-        handleEventSelection={handleEventSelection}
+        handleEventSelection={(selectedEvents) =>
```
kr-matthews, line 23: Sorry if I was unclear, I meant `handleEventSelection={setSelectedEvents}` - no need to create an anonymous function at all.

Frontend/src/pages/registration_administration/components/RegistrationAdministrationList.jsx
```diff
@@ -160,17 +179,14 @@ function RegistrationAdministrationTable({
             <Checkbox
               onChange={(_, data) => {
                 if (data.checked) {
-                  setCheckedBoxes(
-                    registrations.map((registration) => {
-                      add(registration.user.id)
-                      return registration.user.id
-                    })
-                  )
+                  registrations.map((registration) => {
```
kr-matthews, line 182: There's nothing to return here now.
```
registrations.forEach((registration) => add(registration.user.id)
```
In fact could replace the whole block to
```
registrations.forEach((registration) => data.checked ? add(registration.user.id) : remove(registration.user.id)
```

Frontend/src/pages/registration_administration/components/RegistrationAdministrationList.jsx
```diff
@@ -8,6 +8,63 @@ import LoadingMessage from '../../../ui/loadingMessage'
 import styles from './list.module.scss'
 import RegistrationActions from './RegistrationActions'
 
+const reducer = (state, action) => {
+  const { type, attendee } = action
+  switch (type) {
+    case 'add-waiting':
+      return {
+        waiting: [...state.waiting, attendee],
```
kr-matthews, line 16: Nice, looks good overall. Generally reducers are meant to handle complicated state (like we have here) and ensure that someone using the dispatch can't mess up the state.
A potential issue here is that someone could call the dispatch with `add-_` when the attendee is already added. It would be best to check whether the attendee is already in the array, and if so return the _original_ `state` (as opposed to a new copy of the old state, to avoid triggering unnecessary re-renders - though that's probably not significant at this scale).
Another potential issue is that there's nothing stopping someone (future coder, not user) from using the dispatch to add someone from one section (say deleted) into the waiting array. I'm not sure how best to address this (maybe the state should just be a list of selected registrations, and then the 3 sublists are calculated using that and the registrations data? or the selected state should be merged somehow with the registrations state? would have to sit down and think about it some more), but it's probably fine to ignore this for now and revisit it later (but maybe leave a quick comment?).

Frontend/src/pages/register/components/RegistrationEditPanel.jsx
FinnIckler: ah of course, changed.

Frontend/src/pages/registration_administration/components/RegistrationAdministrationList.jsx
FinnIckler: good point. Changed

Frontend/src/pages/registration_administration/components/RegistrationAdministrationList.jsx
FinnIckler: I added a check if the attendee is already in the state, and added a comment about potentially merging states

Frontend/src/pages/registration_administration/components/RegistrationAdministrationList.jsx
```diff
@@ -8,8 +8,18 @@ import LoadingMessage from '../../../ui/loadingMessage'
...
 
+// Currently it is at the developer's discretion to make sure
+// an attendee is added to the right list.
+// One Solution would be to keep the registrations state as
+// the source of truth and partition as needed
 const reducer = (state, action) => {
   const { type, attendee } = action
+  // Make sure no one adds an attendee twice
+  if (
+    [...state.waiting, ...state.accepted, ...state.deleted].includes(attendee)
```
kr-matthews, line 19: Uh won't this block all the `remove-_` actions from removing?

FinnIckler: oops, it was too late yesterday

PR #74, opened 2023-06-19
Add Redis as a Cache

FinnIckler:
After Running the profiler I figured out that updating took so long because it was not caching the calls to the Competitor/Competition Service* when validating the Request. 

This PR adds Redis as a shared cache between the running Handlers.


FinnIckler:
@thewca-bot deploy staging

PR #75, opened 2023-06-23
Implement against Database Test

FinnIckler:
This PR makes some changes to the database structure and implements the database_functions_tests. 

There will be a follow up PR which abstracts stuff like `x["lanes"][0].lane_details["event_details"].map { |event| event["event_id"] }` into more readable code, because that is getting out of hand


dunkOnIT (approved):
LGTM! Thanks for moving the db tests into the correct folder

PR #78, opened 2023-06-23
Add Validations to Create and Update

FinnIckler:
Adds validations for the create and update routes according to the monolith. 

TODO:
- I could extend this PR to include the other routes, but that would require mocking some kind of `user_permissions` endpoint


FinnIckler:
Superseeded by #95

PR #84, opened 2023-06-28
Move Error Codes to their own Class

FinnIckler:
Ruby (in production mode) does not like it if I just define a bunch of constants in a file.


FinnIckler:
@thewca-bot deploy staging

FinnIckler:
@thewca-bot deploy staging

PR #85, opened 2023-06-28
Switch Data Fetching to React-Query

FinnIckler:
This PR adds React-query for data fetching. I also refactored getting the CompetitionInfo to its own Context because it is used on basically every page.
I have not added Error Handling in the Frontend yet, but I have moved all the returned response errors to a custom Error Class that is thrown by `backendFetch` and `externalServiceFetch`


Frontend/src/ui/Competition.jsx
```diff
@@ -0,0 +1,24 @@
...
+  const { data } = useQuery({
+    queryKey: [competition_id],
+    queryFn: () => getCompetitionInfo(competition_id),
+  })
+  const [competitionInfo, setCompetitionInfo] = useState({})
+  useEffect(() => {
+    if (data) {
+      setCompetitionInfo(data)
+    }
+  }, [data])
```
kr-matthews, line 18: You don't need this state or effect, `useQuery` is handling all this for you. Just do `<CompetitionContext.Provider value={{ competitionInfo: data ?? {} }}>` (and can possibly omit the `?? {}` if also doing my comment on the line below).

Typically you won't need any `useEffect`s when working with react-query.

Frontend/src/ui/Competition.jsx
```diff
@@ -0,0 +1,24 @@
...
+  return (
```
kr-matthews, line 19: It would be good to show (only) loading UI while the query is loading, and then guarantee that the competition info being provided to the children is always defined.

Frontend/src/ui/App.jsx
```diff
@@ -11,8 +14,11 @@ export default function App({ children }) {
...
   }, [])
   return (
-    <AuthContext.Provider value={{ user, setUser }}>
-      {children}
-    </AuthContext.Provider>
+    <QueryClientProvider client={queryClient}>
+      <AuthContext.Provider value={{ user, setUser }}>
+        {children}
+      </AuthContext.Provider>
+      <ReactQueryDevtools initialIsOpen={false} />
```
kr-matthews, line 21: What does this do?

Frontend/src/pages/register/components/RegistrationEditPanel.jsx
```diff
@@ -1,27 +1,22 @@
...
+import { CompetitionContext } from '../../../api/helper/context/competition_context'
 import { updateRegistration } from '../../../api/registration/patch/update_registration'
 import { setMessage } from '../../../ui/events/messages'
-import LoadingMessage from '../../../ui/messages/loadingMessage'
 import styles from './panel.module.scss'
 
 export default function RegistrationEditPanel({ registration }) {
   const [selectedEvents, setSelectedEvents] = useState(registration.event_ids)
   const [comment, setComment] = useState(registration.comment)
   const { competition_id } = useParams()
```
kr-matthews, line 13: Do we still need to get this from the params, or is it included in the `competitionInfo`?
Same comment applies to a bunch of other files.

Frontend/src/pages/register/components/RegistrationPanel.jsx
```diff
@@ -3,27 +3,22 @@ import React, { useContext, useState } from 'react'
...
 import styles from './panel.module.scss'
+import { CompetitionContext } from '../../../api/helper/context/competition_context'
 
 export default function RegistrationPanel() {
   const [selectedEvents, setSelectedEvents] = useState([])
   const { competition_id } = useParams()
   const { user } = useContext(AuthContext)
-  const { isLoading, heldEvents } = useHeldEvents(competition_id)
+  const { competitionInfo } = useContext(CompetitionContext)
   const [comment, setComment] = useState('')
```
kr-matthews, line 16: Generally, it would be nice to have the 'global' things - contexts and params - at the top, and 'local' things - states - at the bottom. And a new line before the return.
Same comment applies to a few other components.

Frontend/src/pages/registration_administration/components/RegistrationAdministrationList.jsx
```diff
@@ -100,33 +100,30 @@ const partitionRegistrations = (registrations) => {
...
-  )
+  const { waiting, accepted, deleted } = useMemo(() => {
+    if (registrations) {
+      return partitionRegistrations(registrations)
+    }
+    return {
+      waiting: [],
+      accepted: [],
+      deleted: [],
+    }
```
kr-matthews, line 125: I think `() => partitionRegistrations(registrations ?? [])` would work, though maybe having the still-loading case explicit is nice.

Frontend/src/pages/registration_edit/components/RegistrationEditor.jsx
```diff
@@ -1,35 +1,43 @@
...
+import { CompetitionContext } from '../../../api/helper/context/competition_context'
 import { getSingleRegistration } from '../../../api/registration/get/get_registrations'
 import { updateRegistration } from '../../../api/registration/patch/update_registration'
 import { setMessage } from '../../../ui/events/messages'
 import LoadingMessage from '../../../ui/messages/loadingMessage'
 import styles from './editor.module.scss'
+import { useQuery } from '@tanstack/react-query'
+import getCompetitorInfo from '../../../api/user/get/get_user_info'
 
 export default function RegistrationEditor({ user_id, competition_id }) {
```
kr-matthews, line 13: Can/should we get the user and competition info from the contexts/params like we do everywhere else?

Frontend/src/pages/registration_edit/components/RegistrationEditor.jsx
```diff
@@ -1,35 +1,43 @@
...
+      setComment(serverRegistration.registration.comment)
+      setStatus(serverRegistration.registration.registration_status)
+      setSelectedEvents(serverRegistration.registration.event_ids)
+    }
+  }, [serverRegistration])
 
   return (
     <div className={styles.editor}>
-      {!registration.registration_status || infoLoading || eventsLoading ? (
+      {!registration.registration_status || isLoading ? (
```
kr-matthews, line 40: It would be cleaner to use the `isLoading` from both queries.

Frontend/src/pages/registrations/components/RegistrationList.jsx
```diff
@@ -1,72 +1,72 @@
...
+        `${countrySet.size} Countries`,
+        ...eventCounts.values(),
+        totalEvents,
+      ]
+    }
+    return []
+  }, [registrations, competitionInfo.event_ids])
+  const registrationList = useMemo(() => {
+    if (registrations) {
+      return registrations.map((registration) => {
```
kr-matthews, line 69: Could avoid the explicit `if` and `return`:
```
  useMemo(
    () => registrations?.map((registration) => ...
      ...
    ) ?? [],
    [registrations, competitionInfo.event_ids]
  )
```

Frontend/src/pages/registration_edit/components/RegistrationEditor.jsx
```diff
@@ -1,35 +1,43 @@
...
   const [selectedEvents, setSelectedEvents] = useState([])
+  const [registration, setRegistration] = useState({})
+  const { data: serverRegistration } = useQuery({
+    queryKey: [competition_id, user_id],
+    queryFn: () => getSingleRegistration(user_id, competition_id),
+  })
+  const { isLoading, data: competitorInfo } = useQuery({
+    queryKey: ['info', user_id],
+    queryFn: () => getCompetitorInfo(user_id),
+    enabled: Boolean(serverRegistration),
```
kr-matthews, line 26: `useQuery` provides helpful things like `status` and `isSuccess`, which are designed for things like this (in place of `Boolean(serverRegistration)`).

But why does this query need to be disabled at all? Can't both queries run at once?

One potentially annoying/confusing thing is that `isLoading` from `useQuery` is either `loading`, `success`, or `error`. So if the query is disabled then it's neither a success nor an error so `isLoading` defaults to `true`... even though the query isn't loading.

Frontend/src/pages/registration_edit/components/RegistrationEditor.jsx
```diff
@@ -1,35 +1,43 @@
...
+  })
 
   useEffect(() => {
```
kr-matthews, line 29: `useQuery` takes an `onSuccess` param in the object (and `onError` etc), which would be the cleanest place to put this.
Although maybe that would be triggered every time the query refetches? I forget. If so, then that wouldn't work.
But either way, if the query silently refetches (due to periodic invalidation of cache, or if the page is refocused) and the data has changed in the backend, then this effect is going to set all the state again, overwriting any edits the user was currently making.

Generally, when a query is being used to populate a form like this, my preference is to pass in some params to `useQuery` which prevent it from re-running after the first fetch, and then set all the state in the `onSuccess` param. I can write that out if you want an example.

Frontend/src/pages/registration_administration/components/RegistrationAdministrationList.jsx
```diff
@@ -1,12 +1,12 @@
...
 import { Checkbox, Popup, Table } from 'semantic-ui-react'
-import { getAllRegistrations } from '../../../api/registration/get/get_registrations'
-import getCompetitorInfo from '../../../api/user/get/get_user_info'
 import LoadingMessage from '../../../ui/messages/loadingMessage'
 import styles from './list.module.scss'
 import RegistrationActions from './RegistrationActions'
+import { getAllRegistrations } from '../../../api/registration/get/get_registrations'
 
 // Currently it is at the developer's discretion to make sure
 // an attendee is added to the right list.
```
kr-matthews, line 12: In theory this is slightly more problematic now. `useQuery` will automatically cache query results, and refetch them periodically (and when the page is refocused, etc). (These things can be configured on a query-by-query basis, and possibly globally.) Anyway, if a waiting list person is selected, then the query refetches and that person has since been moved to the accepted list by another organizer on a different computer, then the `selected` state here will still (incorrectly) say that the person is checked on the waiting list, but the UI will (correctly) have updated to show them in the accepted list.

In practice, not sure how much of an issue this will be, and the reducer was going to be revisited later anyway. So probably no need to do anything about it in this PR.

But the general fact that `useQuery` automatically refetches data at periodic/specific times is something to keep in mind, especially in forms.

Frontend/src/ui/App.jsx
FinnIckler: It is supposed to be some handy dev tools, https://tanstack.com/query/v4/docs/react/devtools, but I haven't had any use for them yet...

Frontend/src/ui/Competition.jsx
FinnIckler: ah okay. I removed it and added a `LoadingMessage`

Frontend/src/pages/register/components/RegistrationEditPanel.jsx
FinnIckler: I removed it everywhere except in the first call in `Competition.jsx`

Frontend/src/pages/register/components/RegistrationPanel.jsx
FinnIckler: where does useQuery come in the order? I think otherwise it should be good

Frontend/src/pages/registration_edit/components/RegistrationEditor.jsx
FinnIckler: yeah, changed everywhere

Frontend/src/pages/registration_administration/components/RegistrationAdministrationList.jsx
FinnIckler: Should be fixed by useMutation in the late PR

Frontend/src/pages/registration_edit/components/RegistrationEditor.jsx
FinnIckler: I removed the enabled, I needed it before refactoring something and forgot to remove it.

Frontend/src/pages/registration_edit/components/RegistrationEditor.jsx
FinnIckler: yes, please.

Frontend/src/pages/registration_edit/components/RegistrationEditor.jsx
FinnIckler: I am refactoring this with Mutations and will remove the whole registration state

FinnIckler:
There is a whole second half to this PR where I migrate to useMutation, but will follow up after merging this.

Frontend/src/pages/registration_edit/components/RegistrationEditor.jsx
kr-matthews: Edit: Just realized that the documentation says `onSuccess` is being deprecated in the next major version, so maybe I need to do some more investigating. There might be a better/more standard way. But this is what I was going to suggest:

```
const { ... } = useQuery({
  queryKey: ...,
  queryFn: ...,
  refetchOnWindowFocus: false,
  refetchOnReconnect: false,
  staleTime: Infinity,
  refetchOnMount: 'always',
  onSuccess: populateInputs,
  onError: ...,
});

function populateInputs(serverRegistration) {
  setRegistration(serverRegistration.registration);
  ... // the other 3
}
```
So the data fetches when the page is mounted, but on no other occasions.

Frontend/src/pages/register/components/RegistrationPanel.jsx
kr-matthews: Uh maybe after all the things I already listed?
Obviously the order isn't super important, but I mentioned it because seeing `selectedEvents` and `comment` so far apart felt wrong.

Frontend/src/pages/registrations/components/RegistrationList.jsx
kr-matthews: I don't know about transpiling. I don't think I would call it 'good practice' but I see it a lot. It's fine either way, I personally prefer to keep it shorter and avoid the explicit falsey case when it's not important (the UI using `registrationList` won't show up in the falsey case).

FinnIckler: ok, I will try to use it everywhere where the falsey case is not important

PR #88, opened 2023-07-04
Add guests/administrative comments

FinnIckler:
This PR adds guests/administrative comments as a feature when registering/when administrating registrations. 

In the future we will have a separate guest lane, which allows for guest only registering with payment, but this is not part of this iteration.

### DIFF
--- a/Frontend/src/pages/registration_administration/components/RegistrationAdministrationList.jsx
+++ b/Frontend/src/pages/registration_administration/components/RegistrationAdministrationList.jsx
@@ -175,7 +175,7 @@ function RegistrationAdministrationTable({
   selected,
 }) {
   return (
-    <Table textAlign="left" className={styles.list}>
+    <Table textAlign="left" className={styles.list} singleLine>
       <Table.Header>
         <Table.Row>
           <Table.HeaderCell>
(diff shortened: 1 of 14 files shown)


FinnIckler:
@thewca-bot deploy staging

FinnIckler:
@thewca-bot deploy staging

Frontend/src/pages/registration_administration/components/RegistrationAdministrationList.jsx
```diff
@@ -254,7 +257,24 @@ function RegistrationAdministrationTable({
                   />
                 </Table.Cell>
                 <Table.Cell>{registration.event_ids.length}</Table.Cell>
-                <Table.Cell>{registration.comment}</Table.Cell>
+                <Table.Cell>{registration.guests}</Table.Cell>
+                <Table.Cell title={registration.comment}>
+                  {registration?.comment?.length > 12
+                    ? registration.comment.slice(0, 12) + '...'
+                    : registration.comment}
```
kr-matthews, line 264: I feel like there must be better ways to do this via styling, without changing the actual cell content. And hovering should probably show the whole comment?
If not, then extracting the logic into a helper function would be the next best thing, since it's reused below.

FinnIckler: I tried like 30 minutes yesterday to do it with styling, but Semantic Tables only support it in `fixed singleline` mode, which turns every column into one with ..., even the Country, the name, etc. Hovering already shows the full comment, that's what setting title does!

FinnIckler: I will extract the function and write a comment about the issue with semantic tables

PR #89, opened 2023-07-07
Generate Typescript Types from Swagger

FinnIckler:
This PR adds the pipeline to generate Typescript Types for the backend routes out of the API Spec. With this, the documentation can never be out of sync with the implementation. 
This also gives us the benefits of migrating from the custom `backendFetch` method to the overall typechecked `openapi-fetch` method.
Currently only the /registrations route is done from a testing/documentation point of view, so I only changed one call. With every route done, we can migrate until we can remove backendfetch!


FinnIckler:
Not sure how much there is to check, because this just adds one call and otherwise generates the files, but this would be ready for review.

PR #94, opened 2023-07-17
 Monolith Integration #1: Competition Page

FinnIckler:
Adds a bunch of fields that are now added in the competitions API.
List of Changes:
- Only show competitors tab when competition registration has started
- Correctly show if registration is open/closed
- Show message how long till competition will open / when it closed (using moment)
- Show registration fee in correct decimalization independent from currency (using dinero)
- Respect `allow_registration_edits` and `guests_per_registration_limit`
- Show Custom Organizer defined Tabs (now linkable!)
- Added Competition Home Page
- Link to External Registration Page when external Page is used and don't show the Register Tab


Frontend/src/pages/home/index.jsx
```diff
@@ -1,53 +1,141 @@
...
+              {competitionInfo.venue_address}
+            </li>
+            {competitionInfo.venue_details ? (
+              <li>
+                <span className={styles.detailHeader}>Details: </span>
+                {competitionInfo.venue_details}
+              </li>
+            ) : (
+              ''
+            )}
```
kr-matthews, line 77: There are a few instances of `a ? b : ''` which can be replaced with `a && b`, unless the empty string is important for something.

Frontend/src/pages/register/components/RegistrationPanel.jsx
```diff
@@ -152,7 +162,11 @@ export default function RegistrationPanel() {
             value={guests}
             onChange={(e, data) => setGuests(data.value)}
             selection
-            options={[...new Array(10)].map((_, index) => {
+            options={[
+              ...new Array(
+                (competitionInfo.guests_per_registration_limit ?? 99) + 1 // Arrays start at 0
+              ),
+            ].map((_, index) => {
```
kr-matthews, line 169: I think you can just `options = new Array(...).map(...)`.

Frontend/src/pages/home/index.jsx
FinnIckler: true, I keep on forgetting that...

Frontend/src/pages/register/components/RegistrationPanel.jsx
FinnIckler: you would think so, but otherwise you get a sparse array with 99 empty slots, which you can't map on

PR #95, opened 2023-07-18
Working post test

dunkOnIT:
Merges the testing config that producues a working POST registration test into main.

**NOTE:** The updated request format breaks the db tests - probably want to fix that before merging.


dunkOnIT:
@FinnIckler ready for review :D

FinnIckler (approved):
just changed some minor things so the swagger generated is correct.

PR #96, opened 2023-07-18
Monolith Integration #2: Permission and User Context

FinnIckler:
Now that we have a real Permissions Route and /me Route on the Monolith, we can include them as Contexts as part of the Parent App component.

### DIFF
--- a/Frontend/src/api/schema.d.ts
+++ b/Frontend/src/api/schema.d.ts
@@ -3,120 +3,121 @@
  * Do not make direct changes to the file.
  */
 
+
 export interface paths {
-  '/api/v1/registrations/{competition_id}': {
+  "/api/v1/registrations/{competition_id}": {
     /** List registrations for a given competition_id */
     get: {
       parameters: {
         path: {
-          competition_id: string
-        }
-      }
+          competition_id: string;
+        };
+      };
       responses: {
         /** @description Valid competition_id but no registrations for it */
         200: {
           content: {
-            'application/json': components['schemas']['registration'][]
-          }
-        }
+            "application/json": (components["schemas"]["registration"])[];
+          };
+        };
         /** @description Competition ID doesnt exist */
         404: {
           content: {
-            'application/json': components['schemas']['error_response']
-          }
-        }
+            "application/json": components["schemas"]["error_response"];
+          };
+        };
         /** @description Competition service unavailable - 500 error */
         500: {
           content: {
-            'application/json': components['schemas']['error_response']
-          }
-        }
+            "application/json": components["schemas"]["error_response"];
+          };
+        };
         /** @description Competition service unavailable - 502 error */
         502: {
           content: {
-            'application/json': components['schemas']['error_response']
-          }
-        }
-      }
-    }
-  }
-  '/api/v1/register': {
+            "application/json": components["schemas"]["error_response"];
+          };
+        };
+      };
+    };
+  };
+  "/api/v1/register": {
     /** Add an attendee registration */
     post: {
       parameters: {
         header?: {
-          Authorization?: string
-        }
-      }
+          Authorization?: string;
+        };
+      };
       requestBody: {
         content: {
-          'application/json': components['schemas']['submitRegistrationBody']
-        }
-      }
+          "application/json": components["schemas"]["submitRegistrationBody"];
+        };
+      };
       responses: {
         /** @description only required fields included */
         202: {
           content: {
-            'application/json': components['schemas']['success_response']
-          }
-        }
+            "application/json": components["schemas"]["success_response"];
+          };
+        };
         /** @description user impersonation attempt */
         403: {
           content: {
-            'application/json': components['schemas']['error_response']
-          }
-        }
-      }
-    }
-  }
+            "application/json": components["schemas"]["error_response"];
+          };
+        };
+      };
+    };
+  };
 }
 
-export type webhooks = Record<string, never>
+export type webhooks = Record<string, never>;
 
 export interface components {
   schemas: {
     error_response: {
-      error: number
-    }
+      error: number;
+    };
     success_response: {
-      status: string
-      message: string
-    }
+      status: string;
+      message: string;
+    };
     registration: {
-      user_id: string
-      event_ids: EventId[]
-    }
+      user_id: string;
+      event_ids: (EventId)[];
+    };
     registrationAdmin: {
-      user_id: string
-      event_ids: EventId[]
-      comment?: string
-      admin_comment?: string
-      guests?: number
-    }
+      user_id: string;
+      event_ids: (EventId)[];
+      comment?: string;
+      admin_comment?: string;
+      guests?: number;
+    };
     submitRegistrationBody: {
-      user_id: string
-      competition_id: string
+      user_id: string;
+      competition_id: string;
       competing: {
-        event_ids?: EventId[]
-        comment?: string
-        guests?: number
-      }
-    }
+        event_ids?: (EventId)[];
+        comment?: string;
+        guests?: number;
+      };
+    };
     updateRegistrationBody: {
-      user_id: string
-      event_ids: EventId[]
-      comment?: string
-      admin_comment?: string
-      guests?: number
-    }
-  }
-  responses: never
-  parameters: never
-  requestBodies: never
-  headers: never
-  pathItems: never
+      user_id: string;
+      event_ids: (EventId)[];
+      comment?: string;
+      admin_comment?: string;
+      guests?: number;
+    };
+  };
+  responses: never;
+  parameters: never;
+  requestBodies: never;
+  headers: never;
+  pathItems: never;
 }
 
-export type external = Record<string, never>
+export type external = Record<string, never>;
 
-export type operations = Record<string, never>
+export type operations = Record<string, never>;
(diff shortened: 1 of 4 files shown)


Frontend/src/ui/App.jsx
```diff
@@ -1,23 +1,44 @@
...
-  const setUser = useCallback((user) => {
-    localStorage.setItem(USER_KEY, user)
-    localStorage.removeItem(JWT_KEY)
-    setUserState(user)
-  }, [])
+  const { isLoading, data: user } = useQuery({
+    queryKey: ['user-me'],
+    queryFn: () => getMe(),
+  })
+  const { isLoading: permissionsLoading, data: permissions } = useQuery({
```
coder13, line 22: There should be the "Provider" wrapper component that does this fetch.

You can reference my AuthProvider here: https://github.com/coder13/delegateDashboard/blob/main/src/providers/AuthProvider.js

Frontend/src/ui/App.jsx
```diff
@@ -1,23 +1,44 @@
...
+          value={{
+            permissions,
+            canAttendCompetition: () =>
+              permissions.can_attend_competitions.scope === '*',
+            canAdminCompetition: (id) =>
+              permissions.can_administer_competitions.scope === '*' ||
+              permissions.can_administer_competitions.scope.includes(id),
+          }}
+        >
+          {isLoading || permissionsLoading ? <LoadingMessage /> : children}
```
coder13, line 39: You can expose "loading" from the UserContext and then make a component that renders the app loading the auth if it is.

I also highly recommend caching the auth token / jwt in local storage to make the initial render nicer. I do this for competitiongroups.com.  You can use the expiresAt to determine if the JWT expired. You can also do refresh tokens with JWTs so really as long as someone has a cached JWT which is valid, then there's almost no reason to show a loading stage for user auth.

Frontend/src/api/helper/context/permission_context.ts
```diff
@@ -0,0 +1,14 @@
+import { createContext } from 'react'
+import { Permissions } from '../../auth/get_permissions'
+
+interface PermissionsContext {
+  permissions?: Permissions
+  canAdminCompetition: (id: string) => boolean
+  canAttendCompetition: () => boolean
```
kr-matthews, line 7: Is this meant to be `canAttendCompetitions` (plural)? And below.
Why is it a function instead of just a boolean?

Frontend/src/ui/App.jsx
```diff
@@ -1,23 +1,44 @@
...
+  const { isLoading, data: user } = useQuery({
+    queryKey: ['user-me'],
+    queryFn: () => getMe(),
+  })
+  const { isLoading: permissionsLoading, data: permissions } = useQuery({
+    queryKey: ['permissions-me'],
+    queryFn: () => getPermissions(),
+  })
   return (
     <QueryClientProvider client={queryClient}>
```
kr-matthews, line 27: Does this work? The queries are happening outside the query client provider.

Frontend/src/ui/Tabs.jsx
```diff
@@ -74,7 +75,7 @@ export default function PageTabs() {
         render: () => {},
       },
     ]
-  }, [competitionInfo.id, navigate])
+  }, [canAdminCompetition, competitionInfo.id, navigate])
```
kr-matthews, line 78: I think `canAdminCompetition` will be 'new' every render since it's a function (unless it's memoized somewhere?), so this `useMemo` is running every time.
I'm not sure if adding `canAdminCompetition(competitionInfo.id)` to the dependency array instead is allowed.

Frontend/src/api/helper/context/permission_context.ts
FinnIckler: Currently it could be a boolean, but I am adding conditions to the permissions route soon (e.g if the competitor is banned for competitions before a date) for which it needs to be a function

Frontend/src/ui/App.jsx
FinnIckler: good point, it probably doesn't. I will put it in a Separate File which lives underneath it

Frontend/src/ui/Tabs.jsx
FinnIckler: eslint was the one to warn me about adding it to the dependency array, but I will investigate if it does

Frontend/src/api/helper/context/permission_context.ts
FinnIckler: done

Frontend/src/ui/Tabs.jsx
FinnIckler: not relevant any more now that this is a boolean

FinnIckler:
This is basically done, honestly I didn't really get how to properly wrap it like in @coder13 AuthProvider, because the Auth Provider is still the Monolith and this is just a user data provider

FinnIckler:
the tests are failing because I switched to `https://test-registration.worldcubeassociation.org/api/v10` because I needed the new fields (but the mock targets wca.org), will ask @dunkOnIT what the best course of action is

Frontend/src/index.jsx
```diff
@@ -30,8 +31,10 @@ const router = createBrowserRouter([
         element: (
           <Container>
             <Competition>
-              <PageTabs />
-              <Outlet />
+              <User>
+                <PageTabs />
+                <Outlet />
+              </User>
             </Competition>
```
kr-matthews, line 38: `User` should be something `UserProvider` or something, to make it clear it's just providing the context/data, and not actually some user-based UI which is displaying the children somewhere within it. Probably same with `Competition`.

How come the user provider is inside the competition provider? I would think the user provider would be pretty close to the root, since it sounds like the kind of thing that doesn't change much (only on login/logout), whereas once logged in a user would navigate between multiple competitions.

Frontend/src/ui/User.jsx
```diff
@@ -0,0 +1,53 @@
...
+  })
+  const { isLoading: permissionsLoading, data: permissions } = useQuery({
+    queryKey: ['permissions-me'],
+    queryFn: () => getPermissions(),
+    retry: false,
+    onError: (err) => {
+      setMessage(err.message, 'error')
+    },
+    // Don't try to get permissions if we are not logged in
+    enabled: user.id,
```
kr-matthews, line 30: I believe that when this query is disabled, `permissionsLoading` will be stuck as `true` (even though it's not fetching anything), which might affect the return below?

Frontend/src/index.jsx
FinnIckler: It should be closer to the root, but the permissions context depends on competitions, so it could be Users -> Competitions -> Permissions?

FinnIckler: Currently a user needs to go back to the monolith to change comps so we sadly don't get any benefits there. I will make the users/me route return a 304 though, so at least in most cases it doesn't need to refetch

kr-matthews: Oh, I'm not really familiar with the integration into the monolith.
So is the route `/competitions` handled by the monolith or by wca-registration? Looking at the code for the router I thought it was handled here, but visiting the test site it seems be the monolith.

kr-matthews:
> This is basically done, honestly I didn't really get how to properly wrap it like in @coder13 AuthProvider, because the Auth Provider is still the Monolith and this is just a user data provider

Regardless of which part is handling authentication, you can still have an authentication (or user) context provider wrapper at the root (just inside the query provider wrapper). It provides information about whether the user is logged in, and if so, the user data. And provides functions to login/logout (this part I'm not clear on how it works in this particular case).

This information and functionality shouldn't depend on whether a competition is selected or which competition is selected. So I don't see why it wouldn't be at the root, just inside the query provider.

Currently you're handling the permissions context/provider inside the user provider, which I don't think is the way to go. That should be a separate context, and be inside the competition context.

Something like
```
function App() {
  return (
    <QueryProvider ... >
      <UserProvider>{children}</...>
    </...>
  )
}

// other file
function UserProvider() {
  const userContext = useUser(...);
  return <UserContext.Provider value={userContext}>{children}</...>
}
```
plus analogous files/components for `CompetitionProvider` and `PermissionsProvider`, just like `UserProvider` above.

Then inside the `'/competitions/:competition_id'` branch of the router:
```
element: (
  <CompetitionProvider>
    <PermissionsProvider>
      <PageTabs />
      <Outlet />
    </...>
  </...>
)
```

So the user provider is at the route, and the competition and permissions providers are inside the branch of a specific competition. And note that they are all separate (the permissions provider isn't directly inside the user provider). This is ok because the Permissions provider still has access to the data from the user provider, since the whole app is basically inside the user provider.

Frontend/src/index.jsx
FinnIckler: yeah /competitions is currently handled by the monolith. We do want to move that page to react eventually as part of the competitions service and it would be nice to have some data sharing there, but we see when we get there

FinnIckler: I get the confusing though looking at the router, notice that there is no /competition -> children -> / route!

Frontend/src/ui/User.jsx
FinnIckler: fixed it by doing 
```
return loggedIn && isLoading ? (
    <LoadingMessage />
  )
```
in the new context provider

PR #125, opened 2023-07-25
Monolith Integration PR #3: Payments

FinnIckler:
This is the (still largely untested) integration with the payment service of the monolith implemented against the design we came up with.
There has not been a consensus about polling/event based approaches yet, so I kept it mocked in the implementation. 

TODO:
- [x] Create a test comp far in the future that has payments set up for testing
- [x] Find a way to mock getting the client secret
- [x] Decide on a good way to inject the stripe publishable key into the build (Decided on using a config route)

Polling/Event based state updates will come in the next PR


Frontend/src/pages/register/components/RegistrationPanel.jsx
```diff
@@ -67,191 +69,201 @@ export default function RegistrationPanel() {
...
-            You have registered for {competitionInfo.name}
-          </div>
-          <Divider className={styles.divider} />
-          <div className={styles.registrationRow}>
-            <div className={styles.eventSelectionText}>
-              <div className={styles.eventSelectionHeading}>
-                Your Selected Events:
+    <div>
+      {processing && (
+        <div className={styles.processing}>
```
kr-matthews, line 81: How come this div isn't part of the `Processing` component?

Frontend/src/pages/register/components/Processing.jsx
```diff
@@ -0,0 +1,48 @@
...
+  ) {
+    setProcessing(false)
+  }
+  return (
+    processing && (
+      <div>
+        <div>Your registration is processing...</div>
+        <div>{data && `Registration status: ${data.status.competing}`}</div>
+        <div>
+          {pollCounter > 3 && (
```
kr-matthews, line 32: Would it be more robust to have this in terms of time passed, rather than poll count, in case the frequency is changed in the future?

Frontend/src/pages/register/components/Processing.jsx
```diff
@@ -0,0 +1,48 @@
...
+      data.status.competing === 'incoming')
+  ) {
+    setProcessing(false)
+  }
+  return (
+    processing && (
+      <div>
+        <div>Your registration is processing...</div>
+        <div>{data && `Registration status: ${data.status.competing}`}</div>
+        <div>
```
kr-matthews, line 31: Did you mean to have the `{data &&` type stuff outside each `div`, or is it important that the `div`s show up even if they are empty? (I don't have this running locally so not sure exactly what everything looks like.)

Frontend/src/pages/register/components/RegistrationPanel.jsx
kr-matthews, line 1: The return in here is getting quite big, it would be good to split up into children at some point.

Frontend/src/pages/register/components/StripeWrapper.jsx
kr-matthews, line 1: Is this used anywhere?

FinnIckler: I haven't gotten a test comp set up that uses payments, so this is not used yet. Working on it now!

Frontend/src/pages/register/components/Processing.jsx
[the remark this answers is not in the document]
FinnIckler: yeah, I noticed this too. My current thinking is that we switch processing to be events based, so it will probably need to live closer to the root anyway

Frontend/src/pages/register/components/Processing.jsx
```diff
@@ -0,0 +1,48 @@
...
+      setPollCounter(pollCounter + 1)
+    },
+  })
+  if (
+    data &&
+    (data.status.payment === 'initialized' ||
+      data.status.competing === 'incoming')
+  ) {
+    setProcessing(false)
+  }
```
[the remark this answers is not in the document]
FinnIckler: that's why I was getting that error about setting state incorrectly in the console....

Frontend/src/pages/register/components/Processing.jsx
FinnIckler: I was doing this so I can already start blurring the background except for the Div to indicate that something is processing.

Frontend/src/pages/register/components/Processing.jsx
FinnIckler: good point and time passed still makes sense even if we switch to event based

Frontend/src/pages/register/components/RegistrationPanel.jsx
FinnIckler: do you have any good candidates? I tried thinking about it a couple of weeks ago, when I had to introduce the nth ternary, but I didn't find any obvious candidates

FinnIckler:
This PR is now ready for review. There is still some deployment related things to iron out and I really don't know how we can mock getting a client_secret_id from stripe. Maybe @rubiks-boy could help out?

gregorbg:
There is no way to "mock" a `client_secret_id`. The monolith integration tests use "real" credentials for all Stripe transactions but the "real" account that they are connected to is actually Stripe's test environment.

Feel free to copy the credentials from https://github.com/thewca/worldcubeassociation.org/blob/master/WcaOnRails/.env.test#L5-L7

FinnIckler:
I am already using the credentials for the STRIPE_PUBLISHABLE_KEY, but you are right that I should also use the STRIPE_API_KEY to just generate me an Intent for development.

FinnIckler:
@thewca-bot deploy staging

PR #134, opened 2023-08-01
Monolith Integration #4: Complete Competition Page

FinnIckler:
As all competitor facing tabs should be handled by the new React page I added the events and schedule tabs to it. Currently they are quite simple and the calendar view is not ported over yet. 

TODO:
- [ ] Rebase PR so the Payment Commits are gone
- [ ] Figure out why you only get one day on the schedule view
- [ ] Add button for production that brings you back to the old page to do admin tasks like results posting etc.


Frontend/src/pages/home/index.jsx
```diff
@@ -1,20 +1,25 @@
...
+import React, { useContext, useState } from 'react'
 import { Button } from 'semantic-ui-react'
 import { CompetitionContext } from '../../api/helper/context/competition_context'
-import styles from './home.module.scss'
+import RegistrationRequirements from '../register/components/RegistrationRequirements'
+import styles from './index.module.scss'
 
 export default function HomePage() {
   const { competitionInfo } = useContext(CompetitionContext)
+  const { showAllRequirements, setShowAllRequirements } = useState(false)
```
kr-matthews, line 12: `useState` returns an array, not an object.
I always think `showAllRequirements` is a function when I see it away from the definition, so I'd recommend `[areAllRequirementsShown, setAreAllRequirementsShown]` or something (but feel free to ignore this).

FinnIckler: oops, fixed

PR #187, opened 2023-08-29
Fix Active Tab Index not being Correct on Refresh

FinnIckler:
Currently, when you refresh the page, the active index is not set correctly. I fix this by using react-router's `useLocation` to set the index of the Tabs Component based on the pathname

### DIFF
--- a/Frontend/src/ui/Tabs.jsx
+++ b/Frontend/src/ui/Tabs.jsx
@@ -1,15 +1,39 @@
 import { UiIcon } from '@thewca/wca-components'
 import React, { useContext, useMemo } from 'react'
-import { useNavigate } from 'react-router-dom'
+import { useLocation, useNavigate } from 'react-router-dom'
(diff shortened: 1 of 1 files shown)


kr-matthews:
Not sure if this is in the scope of this PR, but if I type an invalid path name like `https://test-registration.worldcubeassociation.org/competitions/[competition-id-here]/regist` then I get a 404. It would be good to redirect to the general info.

Definitely out of scope, but I'm too lazy to create an issue right now, but if the tab titles are long then they get cut off by the right edge of the page.

FinnIckler:
That would be defined in the routes?

kr-matthews:
> That would be defined in the routes?

Yeah I think so, something like `/competitions/:competition_id/*` at the bottom of the list. I forget exactly how the syntax works (and I'm used to having the routes as children of each other rather than writing out `/competitions/:competition_id/` repeatedly, in which case it's just the last child is `*` and you redirect there).

PR #256, opened 2023-10-12
Competition_api class refactor

dunkOnIT:
Making competition api data more easy to work with


dunkOnIT:
- [ ] move base_comp_url to `comeptition_api_url` in an .env file

FinnIckler:
Maybe switch CompetitionAPI to something more rails-y like this

```
def find(competition_id)
    competition_json = fetch_competition(competition_id)
    if self.competition_exists?
      return CompetitionInfo.new(competition_json)
    end
    nil
  end
  
  def find!(competition_id)
    competition_json = fetch_competition(competition_id)
    if self.competition_exists?
      return CompetitionInfo.new(competition_json)
    end
    raise @error
  end
```

FinnIckler (changes requested):
as per call

app/controllers/registration_controller.rb
```diff
@@ -344,20 +334,20 @@ def is_admin_or_current_user?
...
 
-      competitor_limit = @competition[:competition_info]["competitor_limit"]
-      raise RegistrationError.new(:forbidden, ErrorCodes::COMPETITOR_LIMIT_REACHED) if params["competing"][:status] == 'accepted' && Registration.count > competitor_limit
+      raise RegistrationError.new(:forbidden, ErrorCodes::COMPETITOR_LIMIT_REACHED) if params['competing'][:status] == 'accepted' && Registration.count > @competition.competitor_limit
     end
 
     def get_competition_info!
-      if CompetitionApi.competition_exists?(@competition_id)
-        CompetitionApi.get_competition_info(@competition_id)
+      competition_response = CompetitionApi.new(@competition_id)
```
FinnIckler: Didn't we say we change this to not be an initialized class?

app/controllers/registration_controller.rb
```diff
@@ -286,50 +282,43 @@ def validate_guests!
     end
 
     def guests_valid?
-      @competition[:competition_info]["guest_entry_status"] != "restricted" || @competition[:competition_info]["guests_per_registration_limit"] >= params[:guests]
+      @competition.guest_entry_status != 'restricted' || @competition.guest_limit >= params[:guests]
```
FinnIckler, line 285: I think it makes sense to have this as a guest_allowed method in the competition classe

FinnIckler (approved):
just one nit

app/controllers/registration_controller.rb
```diff
@@ -344,20 +333,11 @@ def is_admin_or_current_user?
...
-      raise RegistrationError.new(:forbidden, ErrorCodes::COMPETITOR_LIMIT_REACHED) if params["competing"][:status] == 'accepted' && Registration.count > competitor_limit
-    end
-
-    def get_competition_info!
-      if CompetitionApi.competition_exists?(@competition_id)
-        CompetitionApi.get_competition_info(@competition_id)
-      else
-        raise RegistrationError.new(:not_found, ErrorCodes::COMPETITION_NOT_FOUND)
-      end
+      raise RegistrationError.new(:forbidden, ErrorCodes::COMPETITOR_LIMIT_REACHED) if params['competing'][:status] == 'accepted' && Registration.count > @competition.competitor_limit
```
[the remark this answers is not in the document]
dunkOnIT: Got #258 open to track it :D

app/controllers/registration_controller.rb
dunkOnIT: Good point! Have changed it to `guest_limit_exceeded`, as that flowed better with the logic of checking for the `:guests` key first

PR #274, opened 2023-10-17
Add Frontend Tests using Cypress

FinnIckler:
When I made the PR to get the Repo up to date I noticed we desperately need Frontend tests to make sure backend and frontend implementation doesn't drift during refactors. This PR adds the following tests:
- Getting a "you have to log in" when trying to register without being logged in
- Registering for a competition after logging in
- Updating registration
- Deleting Registration
- Admin approves a competitor
- Admin edits a competitors registration

I am also fixing some issues that were not noticed in the original PR to make the tests pass


FinnIckler:
fixed #118

Frontend/src/pages/test/login.jsx
```diff
@@ -4,7 +4,11 @@ import { USER_KEY } from '../../ui/providers/UserProvider'
 
 export default function TestLogin() {
   const { login_id } = useParams()
-  useEffect(() => localStorage.setItem(USER_KEY, login_id), [login_id])
+  useEffect(() => {
+    localStorage.setItem(USER_KEY, login_id)
+    // reload the page so the SPA notice that we are 'logged in'
+    window.location = '/'
+  }, [login_id])
```
kr-matthews, line 11: Where is this component used again?
Reloading the page seems like a red flag, even if it's 'just' for a test.

FinnIckler: It is used to "login" to test users by just setting the localstorage instead of setting the user via /users/me. This is because I don't have the localstorage user key as a use effect in the User Context (because it is just for testing)

FinnIckler:
I am merging this to have the bugs that this fixed in the repo, I'll definitely revisit the user management later

PR #282, opened 2023-10-18
Move RegistrationTable back into wca-registrations

FinnIckler:
Not sure why I thought it was a good idea to move a component that is named `registration-table` into wca-components :upside_down_face: .
For some reason data is only correctly displayed on the second fetch now though, can you see something that could cause this @kr-matthews or @coder13 ?

I suspect
 ```
const [state, dispatch] = useReducer(sortReducer, {
    column: '',
    data: registrations,
    direction: undefined,
  })
```

### CODE REVIEW

**Comment by FinnIckler** [comment]
@thewca-bot deploy staging

File: Frontend/src/pages/registrations/components/RegistrationList.jsx [hunk 1]
```diff
@@ -18,15 +65,157 @@ export default function RegistrationList() {
...
+                      )
+                    })}
+                    <Table.Cell>
+                      {registration.competing.event_ids.length}
+                    </Table.Cell>
+                  </Table.Row>
+                ))
+              ) : (
+                <Table.Row>
+                  <Table.Cell width={12}> No matching records found</Table.Cell>
```
kr-matthews, line 201:
```suggestion
                  <Table.Cell width={12}>No matching records found</Table.Cell>
```

Frontend/src/pages/registrations/components/RegistrationList.jsx
```diff
@@ -18,15 +65,157 @@ export default function RegistrationList() {
...
+                </Table.Row>
+              )}
+            </Table.Body>
+            <Table.Footer>
+              <Table.Row>
+                <Table.Cell>{`${newcomers} First-Timers + ${
+                  registrations.length - newcomers
+                } Returners = ${registrations.length} People`}</Table.Cell>
+                <Table.Cell>{`${countrySet.size}  Countries`}</Table.Cell>
```
kr-matthews, line 210: Is this one of those instances of 'country' that need to be changed to 'region' (potentially as a separate PR)?

kr-matthews (commented):
What do you mean by the second fetch?

FinnIckler:
With second fetch I mean something like this. 
Go to Competitors -> Competitors are empty. Go to another tab, then to competitors again -> competitors are there. I guess second render makes more sense

Frontend/src/pages/registrations/components/RegistrationList.jsx
FinnIckler: These changes will all be done as part of integrating i18n

Frontend/src/pages/registrations/components/RegistrationList.jsx
```diff
@@ -18,15 +65,157 @@ export default function RegistrationList() {
...
+                {[...eventCounts.entries()].map(([id, count]) => (
```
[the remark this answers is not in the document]
FinnIckler: no, it returns an [iterator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/entries)

FinnIckler:
@thewca-bot deploy staging

kr-matthews:
> For some reason data is only correctly displayed on the second fetch now though, can you see something that could cause this @kr-matthews or @coder13 ?
> 
> I suspect
> 
> ```
> const [state, dispatch] = useReducer(sortReducer, {
>    column: '',
>    data: registrations,
>    direction: undefined,
>  })
> ```


I've identified the issue (yes, it's the reducer as you suspected) and proposed a couple of solutions on slack.

FinnIckler:
moved data out of the reducer

Frontend/src/pages/registrations/components/RegistrationList.jsx
```diff
@@ -10,40 +10,30 @@ import styles from './list.module.scss'
...
       return {
         ...state,
-        data: state.data.slice().reverse(),
-        direction: state.direction === 'ascending' ? 'descending' : 'ascending',
+        sortDirection:
+          state.sortDirection === 'ascending' ? 'descending' : 'ascending',
       }
     }
-    switch (action.column) {
+    switch (action.sortColumn) {
```
kr-matthews, line 20: All these cases are identical, so all the existing cases can be a single branch of the switch. Or if we're ok removing the "unknown column" error then we could just remove the switch entirely and
```
      return {
        sortColumn: action.sortColumn,
        sortDirection: 'ascending',
      }
```

FinnIckler: I kept the switch and just made the cases fall through.

PR #285, opened 2023-10-19
Refactor tests to use factories instead of fixtures

dunkOnIT:
TODO: 
- [x] create factories
- [x] refactor validations into a service
- [x] check if we have all validations (just take a look for any validation functions missing from controller
- [x] change validate_update_request to use checker service, get old tests running
- [x] go through all old tests, make sure we have the cases covered (add them to a spreadsheet?) 
- [x] once cases are covered, check files for commented out lines and puts statements
- [x] push and flag for review :)


Frontend/src/api/registration/post/submit_registration.ts
```diff
@@ -1,4 +1,4 @@
-import createClient from 'openapi-fetch'
+iport createClient from 'openapi-fetch'
```
gregorbg, line 1: !!?

dunkOnIT:
- Need help resolving the lane_factory conflict
- Just pushing the new stuff for now, will remove the RSwag tests in a later PR

Frontend/src/api/registration/post/submit_registration.ts
dunkOnIT: LMAO

dunkOnIT: We caught this in the previous review of this PR, and I only changed it in the other branch :skull:

dunkOnIT: Should be fixed now

dunkOnIT:
- [ ] Rewrite comeptitor_count function to (a) only consider registrations for one competition, (b) use `where` instead of `all` (consider using a nested query here)
- [ ] Make `target_user_id = @request['user_id'].to_s` as an instance variable in the functions, instead of calling it every time

FinnIckler (approved):
as per call

dunkOnIT:
- [ ] registration_spec.rb should query database to make sure that is_competing status change are persisting - we shouldn't just test the object in memory

PR #286, opened 2023-10-23
Fix more bugs found during testing on staging

FinnIckler:
Had to merge the second PR to continue testing, so here are more fixes that came up when testing


FinnIckler:
@thewca-bot deploy staging

FinnIckler:
@thewca-bot deploy staging

PR #293, opened 2023-10-26
Add Administration Modal

FinnIckler:
Placement is still TBD, but I added this modal to allow access to administration processes through the new UI. 
Might also Group these a bit better

### DIFF
--- a/Frontend/src/ui/Competition.jsx
+++ b/Frontend/src/ui/Competition.jsx
@@ -3,17 +3,20 @@ import { useQuery } from '@tanstack/react-query'
 import { CubingIcon, UiIcon } from '@thewca/wca-components'
 import { dinero, toDecimal } from 'dinero.js'
 import moment from 'moment'
-import React from 'react'
+import React, { useContext, useState } from 'react'
 import { useNavigate, useParams } from 'react-router-dom'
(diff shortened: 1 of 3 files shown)


Frontend/src/ui/Competition.jsx
```diff
@@ -59,22 +62,114 @@ export default function Competition({ children }) {
...
-                  disabled={!competitionInfo['registration_opened?']}
-                  onClick={(_, data) => {
-                    if (!data.disabled) {
-                      if (competitionInfo.use_wca_registration) {
-                        navigate(`${BASE_ROUTE}/${competitionInfo.id}/register`)
-                      } else {
-                        window.location =
-                          competitionInfo.external_registration_page
+                {canAdminCompetitions ? (
+                  <Modal
```
kr-matthews, line 66: This return is already pretty long, it would be good to have a separate component for the admin options.

kr-matthews (commented):
I think the Register button should probably always be present, admin or not.

FinnIckler:
yeah probably, but not sure where else we have space...

FinnIckler:
Closing in favour of using the old sidebar if you are admin

PR #295, opened 2023-10-30
Add Public Waiting List Tab

FinnIckler:
This needs still some way to move people up and down the waiting list so I will continue to work on this after some more pressing manners are done


Frontend/src/pages/waiting/components/WaitingList.jsx
```diff
@@ -0,0 +1,39 @@
...
+    queryKey: ['waiting', competitionInfo.id],
+    queryFn: () => getWaitingCompetitors(competitionInfo.id),
+    retry: false,
+    onError: (err) => {
+      setMessage(err.message, 'error')
+    },
+  })
+  return isLoading ? (
+    <LoadingMessage />
+  ) : (
```
kr-matthews, line 21: Won't this have that problem again of trying to display the waiting list in the error case when `waiting` is `null`?

Frontend/src/pages/waiting/index.jsx
```diff
@@ -0,0 +1,12 @@
+import React from 'react'
+import { Header } from 'semantic-ui-react'
+import WaitingList from './components/WaitingList'
+
+export default function Registrations() {
```
kr-matthews, line 5: Is this meant to be called `Registrations`?

FinnIckler:
@thewca-bot deploy staging

Frontend/src/pages/waiting/components/WaitingList.jsx
kr-matthews, line 1: Thoughts on using "waitlist" everywhere? I think that's quite standard, at least in North America. It also makes all the file and variable names a bit shorter and consistent (sometimes it's `waiting`, sometimes `waiting_list`, etc. at the moment).

Frontend/src/pages/registration_edit/components/RegistrationEditor.jsx
```diff
@@ -151,7 +158,31 @@ export default function RegistrationEditor() {
               checked={status === 'cancelled'}
               onChange={(_, data) => setStatus(data.value)}
             />
+            <br />
+            <Header>Guests</Header>
+            <Input
+              disabled={registrationEditDeadlinePassed}
+              type="number"
+              min={0}
+              max={99}
```
kr-matthews, line 167: Doesn't the competition information include the maximum number of guests allowed (and whether guests are allowed at all)?

Frontend/src/pages/waiting/components/WaitingList.jsx
```diff
@@ -27,12 +27,26 @@ export default function WaitingList() {
         </Table.Row>
       </Table.Header>
       <Table.Body>
-        {waiting.map((w, i) => (
-          <Table.Row key={w.user_id}>
-            <Table.Cell>{w.user_id}</Table.Cell>
-            <Table.Cell>{i}</Table.Cell>
-          </Table.Row>
-        ))}
+        {waiting ? (
```
kr-matthews, line 30: I'm not sure when this would be null vs empty, but presumably both should be handled the same way?
```suggestion
        {waiting?.length ? (
```

dunkOnIT:
Actions: 
- Finn to make waiting list table draggable

dunkOnIT:
## Issues: 
- Moving multiple competitors to waiting list causes duplicate waiting list positions (assuming because they happen simultaneously?)
- Cache is just returning the expiry time, not the cached object?

dunkOnIT:
- [ ] Add redis to the test docker-compose file

PR #296, opened 2023-10-30
Add button to refund Payments to Edit Registrations and show payment status

FinnIckler:
Adds a Button and a Modal to refund payments and shows the payment status


Frontend/src/pages/registration_edit/components/RegistrationEditor.jsx
```diff
@@ -151,8 +151,23 @@ export default function RegistrationEditor() {
...
+                <Button onClick={() => setIsCheckingRefunds(true)}>
+                  {' '}
+                  Show Available Refunds
+                </Button>
+              )}
+            </>
+          )}
+          {isCheckingRefunds && (
+            <Refunds onExit={() => setIsCheckingRefunds(false)} />
+          )}
```
kr-matthews, line 168: Are you doing this rather than `<Refunds isOpen={isCheckingRefunds} onExit={..} />` to avoid the queries in that component from running until open or something like that? I think modals are designed to also be part of the dom, just hidden when closed. One issue I've run into in the past is the open/close animation not working when doing it the way you have it.

Also, what happens if the user closes the modal while the refund is processing? They won't get the 'refund succeeded' message because the component is gone, right? I would consider disabling the ability to close the modal while it's loading.

Frontend/src/pages/registration_edit/components/Refunds.jsx
```diff
@@ -60,7 +65,9 @@ export default function Refunds({ onExit }) {
           ))}
         </Table.Body>
       </Table>
-      <Button onClick={onExit}>Go Back</Button>
+      <Button disabled={isLoading} onClick={onExit}>
```
kr-matthews, line 68: I think the user can still close the modal any time by clicking outside it (I might be misremembering though).

FinnIckler:
@thewca-bot deploy staging

FinnIckler:
@thewca-bot deploy staging

FinnIckler:
@thewca-bot deploy staging

FinnIckler:
@thewca-bot deploy staging

FinnIckler:
@thewca-bot deploy staging

Frontend/src/pages/registration_edit/components/Refunds.jsx
FinnIckler: it doesn't when using the open={prop}

FinnIckler:
@thewca-bot deploy staging

PR #297, opened 2023-10-30
Enforce missing Registration Requirements in the Frontend

FinnIckler:
There were some missing checks in the frontend:
- Not allowing people to register without a comment if comment enforced is set
- Not showing the guest dropdown if the guest count is set to restricted
- Allowing organizers to register early


Frontend/src/pages/register/components/CompetingStep.jsx
```diff
@@ -128,6 +135,12 @@ export default function CompetingStep({ nextStep }) {
           </>
         ) : (
           <>
+            {!competitionInfo['registration_opened?'] && (
+              <Message warning>
+                The Competition is not open yet, you can still register as a
+                competition organizer
```
kr-matthews, line 141:
```suggestion
                Registration is not open yet, but you can still register as a
                competition organizer or delegate.
```

FinnIckler:
Tests are failing because I changed that it to only Organizers and Delegates can sign up early, not admins in general (WST, WRT, board, etc.) I think that is fine, but I'll ask for feedback on slack.

PR #300, opened 2023-10-31
Mock CompetitionInfo and create more Test Competitions

FinnIckler:
Creates some Test Competitions that have edge cases on them. 
I also learned that there are a lot of fields that can be null, something like using_stripe_payments? for example is not true or false, it's true or null....
While adding the favourites test competition I noticed a bug where main_event_id is not set (also did not know this was possible) so I fixed it.


Frontend/src/api/mocks/fixtures.ts
kr-matthews, line 1: Won't these get out-of-date? Ie `NOT_YET_OPEN` has a concrete open date, which will eventually be in the past.

Frontend/src/api/mocks/get_competition_info.ts
```diff
@@ -0,0 +1,39 @@
...
+      return NOT_YET_OPEN
+    }
+    case 'FMCFrance2023': {
+      return COMMENT_REQUIRED
+    }
+    case 'PickeringFavouritesAutumn2023': {
+      return FAVOURITES_COMPETITION
+    }
+    default: {
+      // This allows non mocked response when debugging a certain competition
```
kr-matthews, line 35: The function name is a bit misleading then.

Frontend/src/api/types.d.ts
```diff
@@ -72,21 +74,24 @@ interface CompetitionEvent {
...
-  email: string
-  region: string
-  senior_delegate_id: number
+  location?: string
+  email?: string
+  region?: string
+  senior_delegate_id?: number
   class: string
+  url: string
   //TODO
```
kr-matthews, line 90: Is this still relevant?

Frontend/src/ui/Competition.jsx
```diff
@@ -129,15 +129,17 @@ export default function Competition({ children }) {
...
-                </span>
-              </div>
+              {competitionInfo.main_event_id && (
+                <div>
+                  <span className={styles.eventHeader}>Main Event:</span>
+                  <span className={styles.event}>
+                    <CubingIcon
+                      event={competitionInfo.main_event_id}
+                      selected={true}
+                    />
```
kr-matthews, line 139: We really need a 'non-selectable' version of this.
```suggestion
                    <CubingIcon
                      event={competitionInfo.main_event_id}
                      selected
                    />
```

FinnIckler: It isn't actually selectable by default in my code. That the monolith overwriting my CSS

Frontend/src/api/types.d.ts
FinnIckler: added actual type

Frontend/src/api/mocks/fixtures.ts
FinnIckler: used moment to make sure that won't happen

Frontend/src/api/mocks/get_competition_info.ts
FinnIckler: I would like to keep that functionality though, so we can more easily handle bug reports. Do you have any good idea for a name?

kr-matthews: Maybe add `...WithRealFallback` on to the end, or something to that effect?

FinnIckler: renamed

PR #317, opened 2023-11-07
Use registration_checker service for `create` requests

dunkOnIT:


dunkOnIT:
Change all registration_checker tests to not expect an error, instead of expecting true, and don't return a boolen from and `!` functions

dunkOnIT:
- rename admin_modifying_own_registration to whatever is in the current code (something like organizer_signing_up_themselves)

dunkOnIT:
Closed in favour of #285

PR #339, opened 2023-11-16
added reset endpoint, removed puts statements

dunkOnIT:


dunkOnIT:
- [ ] Invert logic - check `if Rails.env.production?`

FinnIckler (approved):

PR #370, opened 2023-12-04
Proposal for UI revamp

gregorbg:
Key points:
- Visual intuition instead of text
    - For example, instead of writing "You can register for {competition name}", I just put a green, inviting "OPEN" label to the right of the details listing the registration period
    - This also saves us from the hassle of complicated i18n later on, by keeping all text chunks as simple as possible
- Responsive design by relying on SemUI as much as humanly possible
- Avoid custom CSS as much as possible

Gotchas:
- Not very consistent right now on whether Steps or the router parent should be wrapped in `<Segment>`s
- :warning: Might have accidentally deleted some important checks, because I was mostly prototyping by hard-coding "The registration is open" texts without providing alternatives for cases when the registration isn't open
- The registration process itself needs some improvement in layout and display
    - Most notably, the registration "pop up" should be extracted out of the nav component, so that it can be displayed even when you're not navigating the "Register" part but you have an active registration


FinnIckler:
@thewca-bot deploy staging

gregorbg:
@thewca-bot deploy staging

gregorbg:
@thewca-bot deploy staging

PR #377, opened 2023-12-05
Handle no-competitor-limit UI cases

kr-matthews:
Also added some more information to the registration table headers. Though as noted, some kind of sticky/floating box somewhere, which always stays in sight, would probably be a better solution in the future.

This will create more conflicts with the new design PR, since the accordion was completely redone there.

I don't think the codebase has enough blank lines separating chunks of code. :smiley:


FinnIckler:
Actually the new design completely removed the accordion, because we will move it to an extra step in the step panel. Something like 'these are the requirements, do you agree?", so this is definitely all reusable content wise!

FinnIckler:
and thanks for the whitespace gregor definitely agrees haha

FinnIckler (approved):

PR #378, opened 2023-12-06
Refactor Registration Admin selected-rows reducer

kr-matthews:
Avoids storing redundant information in state about which table the selected attendee is in. This technically wasn't causing issues, but future code changes had the possibility of introducing bugs.

Also changed the checkboxes on the header rows to match the state of the rows below (ie is checked if and only if all the rows are checked), and omitted them when their respective table is empty.

Plus I started adding the functionality to hide the changing-status buttons when the user doesn't have that ability - @FinnIckler if you could finish that off that would be great, since you're familiar with who should have what permissions. Or it can be saved for a later ticket.


FinnIckler:
I finished it by adding a isOrganizerOrDelegate to the PermissionsContext

PR #381, opened 2023-12-06
Prevent accepting too many registrations at once (FE)

kr-matthews:
The backend seems to allow this?

Edit: It would probably be good to have a test for this?


FinnIckler:
interesting, I know we have a test for this on the backend. I will try it out and see what's going on

FinnIckler:
After talking with Duncan I think our current implementation of calling up n times from the frontend causes a race condition, so I will introduce a bulk update route in this PR

PR #382, opened 2023-12-07
Add action button to copy emails to clipboard

kr-matthews:
I briefly tried to also add the emails to the payload from the backend, but quickly realized I wasn't entirely sure how to do that and that it's probably best left to someone else. (It would be great if that were added to this PR before merging.)


FinnIckler:
The registration service does not store emails so a new route in the monolith is needed

kr-matthews:
If the backend is going to take some time to get emails working, than I'd rather save it for a separate PR so that I can build on the new `TableRow` component I introduced here.

PR #383, opened 2023-12-08
Hide page tabs 'More' when empty & move registration steps down

kr-matthews:
Sorry for including 2 separate things in 1 PR.

The tabs are still not perfect - they're good on large and small screens but on medium screens they overflow and overlap each other.

Putting the registration steps below the registration info seems logical to me, and also prevents most of the jumping around. The fade transitions we have are nice, and they would be even nicer if the space they take up gradually expanded/contracted, rather than jumping before/after the transition. But I'm not sure how to cleanly do that (only idea is to add a transition to the `min-height` of the container, which kind of requires knowing the height in advance?).

### DIFF
--- a/Frontend/src/pages/register/index.jsx
+++ b/Frontend/src/pages/register/index.jsx
@@ -43,23 +43,6 @@ export default function Register() {
         <div>
           {canAttendCompetition ? (
             <>
-              <Transition
-                visible={showRegisterSteps}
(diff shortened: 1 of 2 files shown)


FinnIckler:
@thewca-bot deploy-staging

FinnIckler:
@thewca-bot deploy staging

PR #385, opened 2023-12-08
Replace 2nd PageTabs with scroll-to-top button

kr-matthews:
The functionality is there, but the styling is not. I'm not sure the cleanest way to have it float in the bottom right corner - probably something with [Sticky](https://react.semantic-ui.com/modules/sticky/)?

If anyone else wants to finish this off, please do so! Otherwise I will come back to it sometime next week.


kr-matthews:
I couldn't figure out how to absolutely position the button without using inline styling.

Desktop:

image

Phone:

image
image
image

FinnIckler (approved):

PR #386, opened 2023-12-09
FE competing step refactor & validation checks

kr-matthews:
Extracted a bunch of conditions/actions as constants, to make the return statements shorter and more understandable. I believe I didn't change the underlying logic, except for always allowing a competitor to delete their registration if it hasn't been accepted yet (previously we were not allowing that if `competitionInfo.allow_registration_self_delete_after_acceptance` was false). But it would be great if someone could double-check that the rest is unchanged (or at least, is what we want).

Minor updates to the previous FE validation logic I had on the registration editor (organizer view), and then more-or-less mirroring that into the register competing step (competitor view).

Unfortunately, the backend doesn't seem to provide info about max events you can register (ie in a favourites competition). I've temporarily hardcoded that as Infinity, we'll need another PR later to fix it.

Fixed the guest total being displayed with a decimal "2.0". Is there any way to update the backend to return a number (an integer ideally) rather than a string? If so, let's do that. If not, it's probably worth updating the registration context to do that translation itself, otherwise any future uses of .guests will continue to need the parseInt wrapper.

Also fixed a bug I introduced when creating `TableRow` - the "edit" link was broken.

I changed the fallback for `competitionInfo.event_change_deadline_date` to `competitionInfo.start_date`, to match what the registration info page says.


FinnIckler:
The weird thing is that the guest issue only turned up once we added guests as a number to the dynamoid model explicitly. I think this is because DynamoDB calls floats numbers. I will change it to integer instead which should fix it?

FinnIckler:
yeah that fixed it

PR #395, opened 2023-12-14
Added check for event limit

dunkOnIT:
Prevents people from registering for more events than the event limit, if it is specified. 

**Dependent on a monolith API update**

(My bad, this should have been `fix/` rather than `feature/`


FinnIckler:
How are you doing this? `events_per_registration_limit` is not in the API reply

dunkOnIT:
Yeah I'm going to PR the monolith to add `events_per_registration_limit` and series competition information, just working on the series comp PR now

dunkOnIT:
- [ ] Change maxEvents on lin 106 of pages/register/etc
- [ ] Add event_registration_limit in types.d.ts

Frontend/src/pages/register/components/CompetingStep.jsx
```diff
@@ -105,7 +105,7 @@ export default function CompetingStep({ nextStep }) {
   const commentIsValid =
     comment.trim() || !competitionInfo.force_comment_in_registration
   // TODO: get max events can register for
```
kr-matthews, line 107:
```suggestion
```

Frontend/src/pages/register/components/CompetingStep.jsx
```diff
@@ -105,7 +105,7 @@ export default function CompetingStep({ nextStep }) {
...
-  const maxEvents = Infinity
+  const maxEvents = competitionInfo.events_per_registration_limit
```
kr-matthews, line 108: This might be necessary?
```suggestion
  const maxEvents = competitionInfo.events_per_registration_limit ?? Infinity
```

Frontend/src/pages/register/components/CompetingStep.jsx
kr-matthews, line 1: `RegistrationEditor.jsx` needs the same changes

dunkOnIT: Done, thanks!

Frontend/src/pages/register/components/CompetingStep.jsx
dunkOnIT: Great catch, thanks!

Frontend/src/pages/register/components/CompetingStep.jsx
dunkOnIT: Removed :heavy_check_mark:

PR #396, opened 2023-12-15
added check for existing reg in series competition

dunkOnIT:
Dependent on a PR in the monolith to add the IDs of all comps in a series to the competition API endpoint


dunkOnIT:
- [ ] Add a series comp to test with frontend (also add an event limit comp) 
- [ ] Update test cases to check for non-cancelled state

kr-matthews:
> * [ ]  Add a series comp to test with frontend (also add an event limit comp)

There is a favourites competitition with an event limit already

dunkOnIT:
- [ ] Add check for existing series reg when updating (to prevent someone re-registering and getting in)

PR #400, opened 2023-12-19
Prefill preferred events from the frontend when registering

FinnIckler:
Adds preferred Events to the user context and makes the default filled in events the subset of the preferred events and the held events


FinnIckler:
@thewca-bot deploy staging

PR #401, opened 2023-12-19
Fix error when not logged in

FinnIckler:
Added two null checks, will check on staging if that was it or if there is more


FinnIckler:
@thewca-bot deploy staging

PR #402, opened 2023-12-20
Handle time zones spilling into other days

kr-matthews:
I continue to hate working with dates in js.

The schedule will include any dates between and including the date of the earliest activity start time and the latest activity start time (from among all activities), **in the selected time zone**. So if you switch to a time zone for which the schedule spills into another day and then you un-select all rooms, the extra day will still appear, but it will just be empty.

See video on slack.


Frontend/src/lib/dates.js
```diff
@@ -1,23 +1,33 @@
...
+    dates.push(nextDate)
+    nextDate = new Date(nextDate)
+    nextDate.setDate(nextDate.getDate() + 1)
   }
-  return range
+  dates.push(nextDate)
+  return dates
+}
+
+export const areOnSameDate = (date1, date2, timeZone) => {
```
FinnIckler, line 17: Why not use moment's isSame function?

kr-matthews: I thought we didn't want to use `moment`. I forget what alternative Gregor recommended, but I can look into that if you want.

FinnIckler: https://moment.github.io/luxon/#/ luxon is what he mentioned

FinnIckler: I will migrate all our moment things to luxon tomorrow or friday, so if you want you can already use it

FinnIckler: don't forget that you have to use `--build ` in the docker compose up statement when adding a new library

kr-matthews: It should be
```
return date1.setZone(timeZone).hasSame(date2.setZone(timeZone), 'day')
```
as long as the parameters are first converted to luxon `DateTime`s.
[`hasSame`](https://moment.github.io/luxon/api-docs/index.html#datetimehassame), [`setZone`](https://moment.github.io/luxon/api-docs/index.html#datetimesetzone)

Might wait until you're done the migration, since I'm a bit busy now anyway. Or you could merge this now and handle this change as part of the migration. (This `areOnSameDate` comparison of dates isn't new to this PR, it was already in the codebase before, just not as a separate function.)

FinnIckler:
I migrated everything from moment to luxon, feel free to rebase your pr and use it

Frontend/src/lib/dates.js
kr-matthews: Done, should be good to go.

PR #404, opened 2023-12-20
Add react-i18next support

FinnIckler:
This PR adds support for internationalization. This by no means adds all translations, neither does it introduce a way to update and deploy the translations. 

I choose to add this now so we can have nicer error messages for testers and can start working on integrations with the current translation workflow.


translations/en.json
```diff
@@ -0,0 +1,28 @@
...
+    "-4003": "The comment provided is too long",
+    "-4004": "Invalid event selection submitted",
+    "-4005": "This competition requires a comment",
+    "-4006": "Competitor limit has been reached",
+    "-4007": "Invalid registration status provided",
+    "-4008": "Registration has closed"
+  },
+  "bookmarks": {
+    "bookmark": "Unbookmarked this competition.",
+    "unbookmark": "You bookmarked this competition. You will get an email 24h before Registration Opens."
```
kr-matthews, line 26: Is this backwards? If not, then `bookmarked` and `unbookmarked` might make more sense.

FinnIckler: yeah I flipped it in the frontend so I also accidentally flipped it in the translation file

PR #405, opened 2023-12-24
Feature/bulk update route

dunkOnIT:
Adds a bulk update route, with corresponding tests, as well as controller tests for the update method. Still needs to be implemented in the frontend. 

Behaviour is as follows:
- Successful request: Returns a JSON containing a mapping of user_id:updated_registration
- Validation failure: Returns a JSON mapping user_id:error_code

No updates will be processed if any validation failures occur.


app/controllers/registration_controller.rb
```diff
@@ -85,51 +86,19 @@ def ensure_lane_exists
...
-        },
-      } }
-    rescue Dynamoid::Errors::Error => e
-      puts e
-      Metrics.registration_dynamodb_errors_counter.increment
-      render json: { error: "Error Updating Registration: #{e.message}" },
-             status: :internal_server_error
-    end
+    render json: { status: 'ok', registration: process_update(params) }
+  rescue StandardError => e
```
FinnIckler, line 90: what happened to rescuing a specific error?

app/services/registration_checker.rb
```diff
@@ -30,6 +30,26 @@ def self.update_registration_allowed!(update_request, competition_info, requesti
...
+    raise RegistrationError.new(:not_found, ErrorCodes::REGISTRATION_NOT_FOUND)
+  end
+
+  def self.bulk_update_allowed!(bulk_update_request, competition_info, requesting_user)
+    raise RegistrationError.new(:unauthorized, ErrorCodes::USER_INSUFFICIENT_PERMISSIONS) unless
+      competition_info.is_organizer_or_delegate?(requesting_user)
+
+    errors = {}
+    bulk_update_request['requests'].each do |update_request|
+      # rubocop:disable Style/RedundantBegin
```
FinnIckler, line 44: why do you need to disable the rubocop error here?

app/controllers/registration_controller.rb
```diff
@@ -148,6 +117,56 @@ def validate_show_registration
...
   end
 
+  def bulk_update
+    updated_registrations = {}
+    update_requests = params[:requests]
+    update_requests.each do |update|
+      updated_registrations[update['user_id']] = process_update(update)
+    end
+
+    render json: updated_registrations
```
FinnIckler, line 127: I would like to return this like { status: ok, updated_registrations: updated_registrations }

app/controllers/registration_controller.rb
dunkOnIT: Good catch, no idea - fixed

app/controllers/registration_controller.rb
dunkOnIT: Fixed

app/services/registration_checker.rb
dunkOnIT: Rubocop wants me to have no `begin` statement and just rescue the error at the bottom of the method - but if I do that:
* I don't have access to the `update_request` for populating the user_id as it is out of the scope of the loop (I think). 
* This means I can't access the user_id to add to the errors hash
* So it seemed easiest to keep the rescue in the scope of the loop by just disabling the rubocop warning

FinnIckler: but it still would be in the scope of the loop right? I think you can just remove the begin

dunkOnIT: You mean like this? 

```ruby
  def self.bulk_update_allowed!(bulk_update_request, competition_info, requesting_user)
    raise RegistrationError.new(:unauthorized, ErrorCodes::USER_INSUFFICIENT_PERMISSIONS) unless
      competition_info.is_organizer_or_delegate?(requesting_user)

    errors = {}
    bulk_update_request['requests'].each do |update_request|
      update_registration_allowed!(update_request, competition_info, requesting_user)
      rescue RegistrationError => e
        errors[update_request['user_id']] = e.error
      end
    end
    raise BulkUpdateError.new(:unprocessable_entity, errors) unless errors == {}
  end
```

If so, that doesn't work, as the syntax isn't valid (there's an extra `end`)

PR #408, opened 2023-12-27
Refactor moment to Luxon

FinnIckler:
Now that moment is deprecated and we are doing more and more with dates I refactored everything that used moment to luxon. Will finish testing everything tomorrow


FinnIckler:
tested

PR #409, opened 2023-12-28
Add qualification restrictions

dunkOnIT:
- [x] What restrictions exist?
- [x] How are they communicated via JSON? 

Create a comp on staging and use the WCIF to check current behaviour.


app/helpers/user_api.rb
```diff
@@ -33,4 +37,12 @@ def self.can_administer?(user_id, competition_id)
     end
     permissions['can_administer_competitions']['scope'] == '*' || permissions['can_administer_competitions']['scope'].include?(competition_id)
   end
+
+  def self.personal_records(user_id)
+    if Rails.env.production?
+      HTTParty.get(permissions_path(user_id), headers: { WCA_API_HEADER => self.get_wca_token })
```
FinnIckler, line 43: Don't think we need a wca_token for this route! Also this still has the permissions_path ;)

dunkOnIT: Ah lol totally forgot to change that line, fixed

FinnIckler:
So currently the qualifications are in the `competition_json`, but you have created a PR that has a separate route for the qualifications, what is your idea?

dunkOnIT:
Oops good save - plan is to generate the qualification data in the competition factory, but then mock it and have the CompetitionInfo object request it from the API - I just missed that part

app/helpers/competition_api.rb
```diff
@@ -122,6 +122,18 @@ def user_can_cancel?
...
+  def qualifications
+    @competition_json['qualifications']
+  end
+
+  def get_qualification_for(event)
+    @competition_json['qualifications'][event]
+  end
+
+  def enforces_qualifications?
+    @competition_json['qualification_results'] == 1 && @competition_json['allow_registration_without_qualification'] == 0
```
FinnIckler, line 134: I'm a bit confused why you use 1 and o instead of true or false?

dunkOnIT: Was looking at the database not the JSON output - fixed

app/helpers/competition_api.rb
dunkOnIT, line 1: @FinnIckler worth taking a look at the changes here, there are basically two changes I was implementing:
1. Have mock_competition as a fallback if fetch_competition doesn't work - but this still doesn't really work because we can't disable_net_connect because of localstack, so fetch_competition may succeed by calling prod
2. Create the ability to either fetch a /competition endpoint, or a child of that endpoint (eg, /competition/:comp_id/qualifications) - but this is perhaps loading too much functionality into the methods?

dunkOnIT:
@FinnIckler the failure is because tests are being run which aren't in my local branch (this is running ~200 tests, this branch only has ~184 - even when I run the `backend-test` container). This isn't a bad thing, but I am curious if you know how this difference comes about? Are we running the main branch's tests as well somehow?

dunkOnIT:
- [ ] Update monolith endpoint to supply date on which PR was achieved
- [ ] Add a date achieved check to the results qualification
- [ ] Understand the difference between `anyResult`, `ranking` and `attemptResult`

app/helpers/competition_api.rb
```diff
@@ -12,27 +12,47 @@ def comp_api_url(competition_id)
...
+    end
+    raise RegistrationError.new(e.http_status, e.error)
   end
 
   # This is how you make a private class method
   class << self
-    def fetch_competition(competition_id)
-      Rails.cache.fetch(competition_id, expires_in: 5.minutes) do
-        response = HTTParty.get(comp_api_url(competition_id))
+    def fetch_competition_data(url_suffix)
```
FinnIckler, line 53: I would just have another route that for each suffix. So fetch_qualifications for example here

app/helpers/competition_api.rb
```diff
@@ -12,27 +12,47 @@ def comp_api_url(competition_id)
...
+      nil
+    else
+      Mocks.mock_competition(url_suffix)
+    end
   end
 
-  def self.find!(competition_id)
-    competition_json = fetch_competition(competition_id)
-    CompetitionInfo.new(competition_json)
+  def self.find!(competition_id, child = nil)
```
FinnIckler, line 33: again, just make a new method here

dunkOnIT:
Ok, coming back to this, there are two issues: 

1. Generating test data
2. Generating test cases

# Test cases
- [ ] Add cases for when the user achieved their PR
- [ ] Make sure cases are added for all qualification types: 
  - [ ] anyResult
  - [ ] ranking
  - [ ] attemptResult
- [ ] Document the behaviour of qualifications, and 
- [ ] Approve documented qualifications with the community (esp. polish organizers who make heavy use of them)

lib/user_api.rb
```diff
@@ -12,14 +12,22 @@ def self.competitor_info_path
     "#{EnvConfig.WCA_HOST}/api/internal/v1/users/competitor-info"
   end
 
+  def self.competitor_qualifications_path(user_id)
+    "#{EnvConfig.WCA_HOST}/api/v0/results/#{user_id}/qualification_data"
+  end
+
   def self.get_permissions(user_id)
-    HTTParty.get(permissions_path(user_id), headers: { WCA_API_HEADER => self.wca_token })
+    JSON.parse(HTTParty.get(permissions_path(user_id), headers: { WCA_API_HEADER => self.wca_token }).body)
```
FinnIckler, line 20: Pretty sure there is a parsed_body in HTTParty

app/services/registration_checker.rb
```diff
@@ -200,5 +216,26 @@ def existing_registration_in_series?
       end
       false
     end
+
+    def competitor_qualifies_for_event?(event, qualification)
+      competitor_qualification_results = UserApi.qualifications(@requestee_user_id)
+      result_type = qualification['resultType']
+
+      competitor_pr = competitor_qualification_results.find { |result| result['eventId'] == event && result['type'] == result_type }
+      return false if competitor_pr.blank?
```
dunkOnIT, line 225: I'm explicitly returning booleans here which feels like a huge code smell, but I haven't been able to come up with something better.

One alternative might be to change this to a `!` method and raise errors instead of returning a boolean?

lib/user_api.rb
```diff
@@ -20,6 +24,10 @@ def self.get_user_info_pii(user_ids)
     HTTParty.post(UserApi.competitor_info_path, headers: { WCA_API_HEADER => self.wca_token }, body: { ids: user_ids.to_a })
   end
 
+  def self.qualifications(user_id)
```
dunkOnIT, line 27: Perhaps this should be named `qualification_data` instead?

spec/controllers/registration_controller_spec.rb
```diff
@@ -1,13 +1,68 @@
 # frozen_string_literal: true
 
 require 'rails_helper'
+require_relative '../support/helper_functions'
 
 describe RegistrationController do
+  describe '#create', :disabled do
```
dunkOnIT, line 7: Note that these tests are disabled - can be re-enabled when #623 passes

spec/factories/request_factory.rb
```diff
@@ -1,7 +1,8 @@
 # frozen_string_literal: true
 
+require_relative '../rails_helper'
```
dunkOnIT, line 3: NB: No idea why I had to change these to require_relative from require - tests were failing with require

spec/spec_helper.rb
```diff
@@ -37,6 +34,8 @@
   # triggering implicit auto-inclusion in groups with matching metadata.
   config.shared_context_metadata_behavior = :apply_to_host_groups
 
+  config.filter_run_excluding disabled: true
```
dunkOnIT, line 37: This allows a :disabled tag to be added to tests

app/services/registration_checker.rb
```diff
@@ -89,6 +91,20 @@ def validate_create_events!
...
     end
 
+    def validate_qualifications!
+      return unless @competition_info.enforces_qualifications?
+      # TODO: Read the request payload in as an object, and handle cases where expected values aren't found
+      event_ids = @request.dig('competing', 'event_ids')
+
+      unqualified_event = event_ids.find do |event|
+        qualification = @competition_info.get_qualification_for(event)
+        next if qualification.nil?
```
gregorbg, line 101: Using `next` inside `find` (or any functional call like `map` etc.) is highly unorthodox.
The purpose of `find` is to skip the entry "on your behalf" so long as you return `false` from the block. So something like `qualification.present? && competitor_qualifies` as a return value should do, eliminating the need for this one line.

app/services/registration_checker.rb
```diff
@@ -89,6 +91,20 @@ def validate_create_events!
...
+        !competitor_qualifies_for_event?(event, qualification)
+      end
+
+      raise RegistrationError.new(:unprocessable_entity, ErrorCodes::QUALIFICATION_NOT_MET) if unqualified_event
```
gregorbg, line 105: Why are you focussing on finding _one_ unqualified event? Wouldn't it make sense to report all unqualified events at once?

Otherwise competitors might go out to compete, think "Yay now I can finally register", come back to their PC, be happy that there's another Worlds spot available for grabs, and be disappointed that there is _yet another_ event they didn't qualify for but the system didn't tell them about previously.

app/services/registration_checker.rb
gregorbg: Well, since boolean is the expected return type of this method, I don't see what's wrong with it.

app/services/registration_checker.rb
```diff
@@ -200,5 +216,26 @@ def existing_registration_in_series?
...
     end
+
+    def competitor_qualifies_for_event?(event, qualification)
+      competitor_qualification_results = UserApi.qualifications(@requestee_user_id)
+      result_type = qualification['resultType']
+
+      competitor_pr = competitor_qualification_results.find { |result| result['eventId'] == event && result['type'] == result_type }
+      return false if competitor_pr.blank?
+
+      return false unless Date.parse(competitor_pr['on_or_before']) <= Date.parse(qualification['whenDate'])
```
gregorbg, line 227: Too much happening in this line. What if parsing fails?

app/services/registration_checker.rb
```diff
@@ -200,5 +216,26 @@ def existing_registration_in_series?
...
+
+      case qualification['type']
+      when 'anyResult', 'ranking'
+        # By this point the user definitely has a result.
+        # Ranking qualifications are enforced when registration closes, so it is effectively an anyResult ranking when registering
+        true
+      when 'attemptResult'
+        competitor_pr['best'].to_i < qualification['level']
```
gregorbg, line 235: Attention! This is not trivial for FMC and MultiBLD. I think you need to properly parse the result instead of just using `to_i` and calling it a day.

lib/competition_api.rb
```diff
@@ -12,7 +12,7 @@ def self.find(competition_id)
     nil
   end
 
-  def self.comp_api_url(competition_id)
+  def self.url(competition_id)
```
gregorbg, line 15: Why this rename?

lib/competition_info.rb
```diff
@@ -6,6 +6,7 @@ class CompetitionInfo
   def initialize(competition_json)
     @competition_json = competition_json
     @competition_id = competition_json['id']
+    @qualifications = fetch_qualifications
```
gregorbg, line 9: Hm, not super happy about making a network call in a constructor.
Are there any other approaches? Why do we even need to hold/cache the `qualifications` data as part of `competition_info`?

Either
- pass in the already fetched thing into the constructor or
- don't have `qualifications` be part of the `competition_info` model and fetch the data from the network on-demand, ie only in places where it's actually needed.

spec/spec_helper.rb
gregorbg: Do we still need this though?
Seems you have moved all potentially disable-able tests out into another PR

dunkOnIT: Correct, but an automatically-applied `disabled` tag can still be useful for local dev I think

app/services/registration_checker.rb
dunkOnIT: The premise here is that the frontend should be communicating with the user on what events they do/don't qualify for - ie, if a user doesn't meet qualification then the event should not be selectable.

Thus, the role of the backend is not to give the user detailed feedback on which qualifications aren't met, it's just to reject the registration outright as soon as it proves to be ineligible. 

I'm open to returning an array of events that the user doesn't meet qualification if you think that would be better, though.

lib/competition_api.rb
dunkOnIT: CompetitionApi.comp_api_url seems redundant - CompetitionApi.url communicates basically the same thing

lib/competition_info.rb
dunkOnIT: My thinking on this is basically: All competition-related information is logically related, and it makes ergonomic and practical* sense to have it be one object, and I want to abstract away the implementation detail of how a comp's qualification data gets sourced from our business logic. 

If we passed the already-fetched thing into the constructor, I still feel like I wouldn't want to be fetching it in our business logic because that just clogs things up further.

_*because we require the qualification info anywhere that we're using a CompetitionInfo object, at least at the moment._



At the moment the qualification data is needed in all places where CompetitionInfo is needed. I think having them as different endpoints makes sense in the monolith, but for our use case in wca-reg, it's much more ergonomic (in my mind at least) to have it all part of one object.

app/services/registration_checker.rb
dunkOnIT: The latest commit declares them as variables before comparison, and I've also added error handling for the case where parsing fails - which I assume is what you wanted? 

It still fails "silently" in the sense that we just return false and it says that qualification isn't met, so a bit of debugging will be needed. At the same time, I don't see a clean way of returning a different error, and both date values come from our API so the input should be fairly consistent (as opposed to user-provided dates, which we'd of course need to parse more carefully)

app/services/registration_checker.rb
dunkOnIT: Disregard, we now return an array of unqualified events in a "data" parameter in the error payload

app/services/registration_checker.rb
dunkOnIT: Addressed

app/services/registration_checker.rb
dunkOnIT: As discussed, straight integer comparison is fine

PR #427, opened 2024-01-14
Shedule selector room colors

kr-matthews:
I copied the util file from the monolith which figures out whether to make the text black or white. Is there a better way to use the same helper function?

image

image

Max 4 per row:

image

Hidden when the selected venue only has 1 room:

image

Narrow screen:

image

Narrower screen:

image


FinnIckler:
Currently no way to share code like that between the monolith and the registration service

PR #428, opened 2024-01-15
Add internal registrations route

FinnIckler:
Adds a route so the monolith can get the registrations


gregorbg:
@thewca-bot deploy staging

PR #431, opened 2024-01-16
Use the new field to get accepted competitor

FinnIckler:


FinnIckler:
@thewca-bot deploy staging

PR #461, opened 2024-02-16
Cache userData in the Backend and add useUserData hook

FinnIckler:
Instead of Fetching userData seperately, I added a useUserData hook. This user data is then cached in the backend


Frontend/src/pages/waiting/components/WaitingList.jsx
```diff
@@ -24,7 +25,22 @@ export default function WaitingList() {
...
+  )
+
+  const registrationsWithUser = useMemo(() => {
+    if (waiting && userInfo) {
+      return waiting.map((r) => {
+        r.user = userInfo.find((u) => u.id === r.user_id)
+        return r
+      })
+    }
+    return []
```
kr-matthews, line 40: This is repeated at least 3 times, a util function would probably be good.

Frontend/src/hooks/useUserData.ts
```diff
@@ -0,0 +1,14 @@
+import { useQuery } from '@tanstack/react-query'
+import { getCompetitorsInfo } from '../api/user/post/get_user_info'
+
+export function useUserData(ids: number[]) {
+  return useQuery({
+    queryFn: () => (ids.length > 0 ? getCompetitorsInfo(ids) : []),
```
kr-matthews, line 6: Should `getCompetitorsInfo` have `Users` rather than `Competitors` in the name, for consistency with everything else?
And for the hook, I can't decide whether `useUsersData` would be a better name, since it accepts multiple ids.

Frontend/src/api/schema.d.ts
```diff
@@ -5,6 +5,26 @@
...
+    post: {
+      requestBody: {
+        content: {
+          "application/json": {
+            ids: components["schemas"]["userIds"];
+          };
+        };
+      };
+      responses: {
+        /** @description Successfully passed down the Psych Sheet */
```
gregorbg, line 19: copy-paste whoopsie?

Frontend/src/hooks/useUserData.ts
gregorbg, line 6: I think `getCompetitorsInfo` should be able to deal with empty arrays without much hassle.

gregorbg: After tracing the call, it seems like passing an empty array here to `getCompetitorsInfo` would post an empty array to the internal `api/v1`, which goes to the Rails `user_controller#info`. And in the body of that method, the (potentially empty!) IDs are being passed to `RedisHelper.cache_info_by_ids`.

IMHO, that `cache_info_by_ids` method should be able to deal with empty lists and just casually give back an empty list.

gregorbg: +1, especially because you're using `user-info` as `queryKey` in the Tanstack-Query config a few lines down below

Frontend/src/hooks/useUserData.ts
```diff
@@ -0,0 +1,14 @@
...
+    queryKey: ['user-info', ...ids],
```
gregorbg, line 7: Super nitpick, but it might be crucial to actually sort the IDs. We have data tables with custom sorting order, so it is a very realistic use-case for our registration code to sometimes call this as `useUserData([3,5,19])` and sometimes as `useUserData([5,3,19])`. Both yield the same result in terms of the underlying data, but currently would not be cached efficiently by Tanstack-Query.

app/helpers/mocks.rb
```diff
@@ -12,6 +12,39 @@ def self.pii_mock(user_ids)
...
+        {
+          'id' => u,
+          'created_at' => Time.now.to_s,
+          'updated_at' => Time.now.to_s,
+          'name' => "Name #{u}",
+          'wca_id' => "2023TEST#{u % 99}",
+          'delegate_status' => nil,
+          'gender' => 'm',
+          'country_iso2' => iso,
+          'url' => "https://#{EnvConfig.WCA_HOST}/persons/2023TEST#{u % 99}",
```
gregorbg, line 28: Why are you not refactoring the WCA ID into a variable if it's used in two places?

swagger/v1/swagger.yaml
```diff
@@ -127,6 +163,28 @@ components:
...
+            schema:
+              type: object
+              properties:
+                ids:
+                  "$ref": "#/components/schemas/userIds"
+              required: [ids]
+        required: true
+      responses:
+        '200':
+          description: "Successfully passed down the Psych Sheet"
```
gregorbg, line 181: Copy-paste whoopsie?

gregorbg (changes requested):

Frontend/src/hooks/useUserData.ts
```diff
@@ -0,0 +1,15 @@
+import { useQuery } from '@tanstack/react-query'
+import { getUsersInfo } from '../api/user/post/get_user_info'
+
+export function useUserData(ids: number[]) {
+  const sortedIds = ids.sort((a, b) => a - b)
```
gregorbg, line 5: Why not just `ids.toSorted()`? Why do you even pass in a custom comparator?

Frontend/src/api/user/post/get_user_info.ts
```diff
@@ -0,0 +1,28 @@
+import createClient from 'openapi-fetch'
+import { BackendError } from '../../helper/backend_fetch'
+import { components, paths } from '../../schema'
+
+const { POST } = createClient<paths>({
+  // TODO: Change this once we are fully migrated from backend fetch
```
gregorbg, line 6: Why is this needed right now?

app/lib/redis_helper.rb
```diff
@@ -19,6 +19,7 @@ def self.decrement_or_initialize(key, &block)
 
   # The Hydration block needs to return an array of hashes that have an 'id' field
   def self.cache_info_by_ids(key_prefix, ids, &hydration_block)
+    return [] if ids.empty?
```
gregorbg, line 22: This is getting close to what I want, but why is this explicit check even necessary in the first place?

If the user happens to pass in an empty array of ids, then the `keys` variable below maps over an empty array, which isn't forbidden. It simply returns an empty array.

That empty array is passed to `Rails.cache.read_multi`, can it deal with empty arrays? Ideally, I would just expect that method to return an empty hash into `info` as well. Then, `uncached_ids` will be empty, the `unless` block never gets entered, and `info.values` happily returns an empty array.

What am I missing here?

Frontend/src/hooks/useUserData.ts
FinnIckler: toSorted() does not exist. I added the compare function because the typescript linter wants it

gregorbg: Huh??
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/toSorted

Frontend/src/api/user/post/get_user_info.ts
FinnIckler: There's still a couple of methods that use backend fetch, because they haven't been added to the swagger definition. Definitely something we should do very soon!
Backendfetch requires the API URL to be the full URL (+ the /api/v1), while createClient doesn't

app/lib/redis_helper.rb
FinnIckler: It didn't work for some reason 2 days ago, but now I just removed it and it just works, so not sure what happened

Frontend/src/hooks/useUserData.ts
FinnIckler: That function is part of ESNext.Array (from 2023), which is not enabled in our typescript compiler options. I could add it?

gregorbg: Yup, although according to https://stackoverflow.com/a/77569768 the newer TypeScript releases (5.2 and after) support it natively.

I couldn't find the place where we pin our TS version, as it doesn't seem to be in package.json. Can you shed some light on this?

FinnIckler: I never explicitly added it because we only use it for the library functions. It's just a transitory dependency from openapi-typescript and set to "version": "5.1.6". I'll add it explicitly

gregorbg (approved):

PR #462, opened 2024-02-16
Show Minutes and Seconds in Cutoff

FinnIckler:


Frontend/src/pages/events/index.jsx
```diff
@@ -16,6 +16,15 @@ import { CompetitionContext } from '../../api/helper/context/competition_context
 import { setMessage } from '../../ui/events/messages'
 import LoadingMessage from '../../ui/messages/loadingMessage'
 
+function centiSecondsToHumanReadable(centiSeconds) {
```
gregorbg, line 19: I feel like we have a similar function over in the Monolith already. Did you copy this method from over there?

FinnIckler: There's two functions. 
https://github.com/thewca/worldcubeassociation.org/blob/d3338d0175c608eb34bc1d3be60034cf0ac44e46/WcaOnRails/app/webpacker/lib/wca-live/attempts.js#L150
and https://github.com/thewca/worldcubeassociation.org/blob/d3338d0175c608eb34bc1d3be60034cf0ac44e46/WcaOnRails/app/webpacker/lib/utils/edit-events.js#L75
I'll see if these are correct

gregorbg: `centisecondsToClockFormat` does not include any humanized strings like `"seconds"` and instead just returns the raw parsed time. You probably want to go with the one from `edit-events`.

Sidenote: Do we have a plan in mind for sharing frontend logic? Further down the line, there will be many more cases of "stealing" JS code from the monolith

FinnIckler:
We currently do not have a strategy in place for sharing code. I would think an npm package using typescript would be the best idea.

gregorbg:
> We currently do not have a strategy in place for sharing code. I would think an npm package using typescript would be the best idea.

What about https://github.com/thewca/wca-helpers?

gregorbg (commented):
Changes look okay, but does this consider 333fm and 333mbld correctly?

FinnIckler:
It does not. For mbld it uses attemptResult key instead. I'll add it now

FinnIckler:
There is actually already a format centisecond method in wca-helpers https://github.com/thewca/wca-helpers/blob/master/src/helpers/time.ts, but it doesn't include the string. 
I'll add a PR that adds these options and includes the other methods

FinnIckler:
That's done https://github.com/thewca/wca-helpers/pull/30

FinnIckler:
I'm going to merge this for now and then just make another PR when we update wca-helpers

PR #463, opened 2024-02-17
Add internal route to get all registrations for a user

FinnIckler:
For the my-competitions page on the monolith

### DIFF
--- a/app/controllers/internal_controller.rb
+++ b/app/controllers/internal_controller.rb
@@ -62,4 +62,10 @@ class InternalController < ApplicationController
 
     render json: registrations
   end
+
+  def registrations_for_user
+    user_id = params.require(:id)
+    registrations = Registration.where(user_id: user_id).map { |x| { competition_id: x.competition_id, status: x.competing_status } }
+    render json: { registrations: registrations }
+  end
 end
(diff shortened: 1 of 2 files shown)


config/routes.rb
```diff
@@ -12,6 +12,7 @@
   get '/healthcheck', to: 'healthcheck#index'
   post '/api/internal/v1/update_payment', to: 'internal#update_payment_status'
   get '/api/internal/v1/:competition_id/registrations', to: 'internal#list_registrations'
+  get '/api/internal/v1/users/:user_id/registrations', to: 'internal#list_registrations'
```
gregorbg, line 15: copy-paste whoopsie in the `to:` method name?

app/controllers/internal_controller.rb
```diff
@@ -62,4 +62,10 @@ def list_registrations
 
     render json: registrations
   end
+
+  def registrations_for_user
+    user_id = params.require(:user_id)
+    registrations = Registration.where(user_id: user_id).map { |x| { competition_id: x.competition_id, status: x.competing_status } }
+    render json: { registrations: registrations }
```
gregorbg, line 69: Is it necessary to have this top-level `registrations` key? The other method that lists registrations for a competition just does `return json: registrations` directly, and I feel like that's the more concise way to do it.

gregorbg (changes requested):

app/controllers/internal_controller.rb
FinnIckler: good point, I implemented it without that key on the monolith as well

PR #480, opened 2024-02-24
Update Node to 20 in production/staging

FinnIckler:
Updates the Dockerfiles to use Node 20


FinnIckler:
Treating as a dependency update (we already run node 20 in dev no problem)

PR #483, opened 2024-02-25
Make RegistrationAdministrationList sortable

FinnIckler:
Also automatically sorts by registered_on or paid_on, by default.

Sorting order actually doesn't change currently. I'll figure that out tomorrow

### DIFF
--- a/Frontend/src/pages/registration_administration/components/RegistrationAdministrationList.jsx
+++ b/Frontend/src/pages/registration_administration/components/RegistrationAdministrationList.jsx
@@ -10,12 +10,12 @@ import { getAllRegistrations } from '../../../api/registration/get/get_registrat
 import { useUserData } from '../../../hooks/useUserData'
 import { getShortDateString, getShortTimeString } from '../../../lib/dates'
 import { addUserData } from '../../../lib/users'
+import { createSortReducer } from '../../../reducers/sortReducer'
 import { BASE_ROUTE } from '../../../routes'
 import { setMessage } from '../../../ui/events/messages'
 import LoadingMessage from '../../../ui/messages/loadingMessage'
 import styles from './list.module.scss'
 import RegistrationActions from './RegistrationActions'
-import { createSortReducer } from '../../../reducers/sortReducer'
 
 const selectedReducer = (state, action) => {
   let newState = [...state]
@@ -53,6 +53,8 @@ const sortReducer = createSortReducer([
   'events',
   'guests',
   'paid_on',
+  'comment',
+  'dob',
 ])
 
 const partitionRegistrations = (registrations) => {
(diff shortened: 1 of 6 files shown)


Frontend/src/pages/registration_administration/components/RegistrationAdministrationList.jsx
```diff
@@ -44,6 +45,16 @@ const selectedReducer = (state, action) => {
   return newState
 }
 
+const sortReducer = createSortReducer([
```
kr-matthews, line 48: People like to sort by comment, for example for a multi-location FM competition where you need to put your location in the comment.
And what about DOB? I'm not sure how useful that is, but it's possible in the existing registration.

FinnIckler:
I finally added some seed users and sorting works on all columns

Frontend/src/pages/registration_administration/components/RegistrationAdministrationList.jsx
```diff
@@ -136,9 +159,47 @@ export default function RegistrationAdministrationList() {
...
+    return []
+  }, [registrationsWithUser, sortColumn, sortDirection])
+
   const { waiting, accepted, cancelled, pending } = useMemo(
-    () => partitionRegistrations(registrationsWithUser ?? []),
-    [registrationsWithUser],
+    () => partitionRegistrations(sortedRegistrationWithUser ?? []),
+    // this useMemo will not fire again if we only change the order of the array, which is why we need the sortColumn and sortDirection in the deps
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+    [sortedRegistrationWithUser, sortColumn, sortDirection],
```
gregorbg, line 202: I don't get this. `sortedRegistrationWithUser` is itself a `useMemo`, whose dependencies contain `sortColumn` and `sortDirection`.

You are arguing that this current memo needs to trigger if sort order changes, but when that happens the memo of `sortedRegistrationsWithUser` triggers, and by transitive rule this memo should then also trigger. So from what I can tell by staring at the code, your extra deps are totally unnecessary.

There has to be a cleaner solution than overriding good React principles and standards.

gregorbg: Wow, thanks for explaining, that would probably have taken me half an hour to figure out :smile: 
I'm tempted to start a rant why on earth JS sorts in-place, but I guess the short answer is "legacy reasons"...

Made a PR over at https://github.com/thewca/wca-registration/pull/490

kr-matthews: Half an hour wouldn't be so bad; it took me several months to understand why my code didn't work as expected when I first started React...

PR #488, opened 2024-02-26
Cleanup Swagger and remove backend fetch

FinnIckler:
Another refactor that had a long time coming


FinnIckler:
Merging this so I can add contract tests

PR #490, opened 2024-02-27
Use toSorted and toReversed instead of in-place array modifications

gregorbg:


FinnIckler (approved):
Thanks for the cleaner version

PR #507, opened 2024-03-05
Add rest of internationalisation

FinnIckler:
Still compiling what keys need to be added, so the en.json is currently WIP


Frontend/src/pages/registration_edit/components/Refunds.jsx
```diff
@@ -85,7 +89,7 @@ export default function Refunds({ open, onExit, userId, competitionId }) {
         </Modal.Content>
         <Modal.Actions>
           <Button disabled={isMutating} onClick={onExit}>
-            Go Back
+            {t('en.competitions.registration_v2.refunds.back')}
```
kr-matthews, line 92: Is this meant to start with `en.`?

Frontend/src/pages/registration_edit/components/Refunds.jsx
```diff
@@ -42,7 +46,7 @@ export default function Refunds({ open, onExit, userId, competitionId }) {
         <Modal.Content>
           <Table>
             <Table.Header>
-              <Table.Header> Amount </Table.Header>
+              <Table.Header>{t('')}</Table.Header>
```
kr-matthews, line 49: This is empty.

FinnIckler: thanks. As this is organizer only, this won't translated

PR #515, opened 2024-03-13
Fix two broken messages

FinnIckler:
- Don't show two messages when registering
- Don't throw an error in the backend when a registration can't be found


gregorbg:
Just for context: Why are you removing the `status: :not_found` in the backend?
Seems like a reasonable and good design to have the data API return HTTP response 404 if something isn't found...

FinnIckler:
because otherwise the fetch throws an error in the console, which is kinda ugly and confusing for users. Another solution would be to just ignore 404 errors in the fetch

PR #519, opened 2024-03-22
Add Registration History

FinnIckler:
Saves the Registration History in another table and adds another entry on update


Frontend/src/pages/registration_edit/components/RegistrationEditor.jsx
```diff
@@ -287,6 +288,31 @@ export default function RegistrationEditor() {
...
+              <Table.Row>
+                <Table.HeaderCell>Timestamp</Table.HeaderCell>
+                <Table.HeaderCell>Changes</Table.HeaderCell>
+                <Table.HeaderCell>Acting User</Table.HeaderCell>
+              </Table.Row>
+            </Table.Header>
+            <Table.Body>
+              {registration.history.map((entry) => (
+                <Table.Row key={entry.timestamp}>
+                  <Table.Cell>{getShortDateString(entry.timestamp)}</Table.Cell>
```
kr-matthews, line 303: For consistency with other uses of short date strings in tables:
```suggestion
      <Table.Cell>
        <Popup
          content={getShortTimeString(registered_on)}
          trigger={<span>{getShortDateString(registered_on)}</span>}
        />
      </Table.Cell>
```

FinnIckler:
Added an action column:
image
The "Competitor update" to accepted is a bit confusing because the competitor is also the admin, for other cases it would show "Admin Update" instead.

FinnIckler:
Added a test to confirm that the history is populated on update

app/controllers/registration_controller.rb
```diff
@@ -274,6 +274,26 @@ def import
 
   private
 
+    def action_type(request)
+      self_updating = request[:user_id] == @current_user
+      status = request.dig('competing', 'status')
+      if status == 'cancelled'
+        return self_updating ? 'Competitor delete' : 'Admin delete'
+      end
+      self_updating ? 'Competitor update' : 'Admin update'
```
gregorbg, line 283: For future-proofness (especially when we get automated acceptance which means more action_types to take into account) you should probably translate these to integer codes and have i18n translations for the codes, similar to how we already handle some error messages.

Frontend/src/pages/registration_edit/components/RegistrationEditor.jsx
```diff
@@ -40,17 +52,22 @@ export default function RegistrationEditor() {
...
   })
 
-  const { isLoading, data: competitorInfo } = useQuery({
+  const { isLoading, data: competitorsInfo } = useQuery({
     queryKey: ['info', userId],
-    queryFn: () => getUserInfo(userId),
+    queryFn: () =>
+      getUsersInfo([
+        userId,
+        ...serverRegistration.history.map((e) => e.acting_user_id),
```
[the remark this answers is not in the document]
gregorbg: Also, don't we have a custom `useUserData` hydration hook for this?!

app/worker/registration_processor.rb
```diff
@@ -41,6 +44,7 @@ def lane_init(competition_id, user_id)
       empty_registration = Registration.new(attendee_id: "#{competition_id}-#{user_id}",
                                             competition_id: competition_id,
                                             user_id: user_id)
+      RegistrationHistory.create(attendee_id: "#{competition_id}-#{user_id}", entries: [History.new({ 'changed_attributes' => {}, 'actor_user_id' => user_id, 'action' => 'Worker create' })])
```
gregorbg, line 47: Too much happening on one line. Please split the `History.new` thing into its own line.

gregorbg (changes requested):

app/controllers/registration_controller.rb
gregorbg: (feel free to push this to a follow-up PR if you feel that it's too much work)

Frontend/src/pages/registration_edit/components/RegistrationEditor.jsx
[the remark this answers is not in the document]
FinnIckler: we do, but it only works on objects with user_id as a field

Frontend/src/pages/registration_edit/components/RegistrationEditor.jsx
```diff
@@ -287,6 +310,49 @@ export default function RegistrationEditor() {
               />
             </>
           )}
+          <Header>Registration History</Header>
```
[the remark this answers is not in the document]
FinnIckler: that's easy

app/controllers/registration_controller.rb
FinnIckler: I'll do it in another PR because it also requires new I18n strings

gregorbg (approved):

PR #520, opened 2024-03-25
Minor frontend fixes

FinnIckler:
I think there might be some more minor phone fixes needed, I'll add them to this PR

### DIFF
--- a/Frontend/src/overrides.css
+++ b/Frontend/src/overrides.css
@@ -3,6 +3,7 @@
     padding-left: 0 !important;
     padding-right: 0 !important;
 }
+
 /* From application.css */
 .container.fluid {
   padding-right:15px;
--- a/Frontend/src/overrides.css
+++ b/Frontend/src/overrides.css
@@ -1,5 +1,5 @@
 /* We don't want to have the padding that is caused by having two containers in admin mode */
-.container-fluid > #registration-app > .registration-container {
+.container-fluid .registration-container {
     padding-left: 0 !important;
     padding-right: 0 !important;
 }
--- a/Frontend/src/overrides.css
+++ b/Frontend/src/overrides.css
@@ -1,7 +1,14 @@
 /* We don't want to have the padding that is caused by having two containers in admin mode */
-.container > #registration-app > .registration-container {
+.container-fluid > #registration-app > .registration-container {
     padding-left: 0 !important;
     padding-right: 0 !important;
+}
+/* From application.css */
+.container.fluid {
+  padding-right:15px;
+  padding-left:15px;
+  margin-right:auto;
+  margin-left:auto
 }
 
 .competition-info-logo {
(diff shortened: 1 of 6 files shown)


Frontend/src/ui/messages/flash.module.scss
```diff
@@ -1,5 +1,12 @@
 @import '../../style/font-sizes';
 
+@media screen and (max-width: 400px) {
```
gregorbg, line 3: Not a big fan of introducing our own hard-coded breakpoints.
Is the `400px` documented somewhere, or can you reuse breakpoints from SemUI?

Frontend/src/overrides.css
```diff
@@ -1,5 +1,9 @@
-/* We don't want to have the padding that is caused by having two containers */
-.registration-container {
+/* We don't want to have the padding that is caused by having two containers in admin mode */
+.container > #registration-app > .registration-container {
     padding-left: 0 !important;
     padding-right: 0 !important;
+}
+
+.competition-info-logo {
+    margin-top: 0 !important;
```
gregorbg, line 8: Are you sure it's the *top* margin that needs to be fixed?
In my dev console at first glance it looks like the *bottom* margin is causing problems, which makes sense because the clickable icons on the bottom of the comp image appear shifted.

FinnIckler: You would think so, but look at this 
margin-enabled
margin fixed

Frontend/src/ui/messages/flash.module.scss
FinnIckler: sure, I'll find them out

Frontend/src/overrides.css
gregorbg: Ah, that's because in this _specifc_ case, removing the top margin shifts the image upwards enough to make sure that the bottom margin doesn't matter anymore.

I'm not sure whether that's a clean solution. The bottom margin is clearly the one causing the issue.
Can you ask ChatGPT about this? (In two vertically adjacent `div`s, there are elements floated to the right. The top one has... The bottom one becomes misaligned because... Should you fix this by removing the top row's content's margin-top or margin-bottom?)

Frontend/src/overrides.css
```diff
@@ -1,5 +1,16 @@
-/* We don't want to have the padding that is caused by having two containers */
-.registration-container {
+/* We don't want to have the padding that is caused by having two containers in admin mode */
+.container-fluid > #registration-app > .registration-container {
```
gregorbg, line 2: Where is the class `.container-fluid` coming from?
I always thought that `.container` and `.fluid` are two separate classes

FinnIckler: It's from the monolith. It's because in admin mode there is an extra container-fluid from the monolith's application css

gregorbg: Ah, yikes!
Well I guess as long as it works... Monolith UI is deprecated so I don't want to touch that code over there too much :(

gregorbg (approved):

PR #575, opened 2024-05-01
Remove Frontend from Repo

FinnIckler:
Frontend is now served by the monolith directly. Also removes backend routes and mocks that are not needed anymore, now that this has to be run alongside of the monolith


FinnIckler:
yayyyy tests are finally passing. This was needed for ongoing development which requires running both the monolith and the registration system

PR #588, opened 2024-05-21
Return whole registration object in my_registrations

FinnIckler:
This should fix the My Registration on the monolith and aligns it with the data returned from the other routes


FinnIckler:
@thewca-bot deploy staging

PR #589, opened 2024-05-24
Allow a banned user to register for a competition which starts after their ban ends

dunkOnIT:
This address an issue found in QA on the Register page

### DIFF
--- a/app/helpers/user_api.rb
+++ b/app/helpers/user_api.rb
@@ -34,17 +34,10 @@ class UserApi < WcaApi
     end
 
     competition_permissions = permissions['can_attend_competitions']
-
-    # User can compete if they have global attend permissions
-    return true if competition_permissions['scope'] == '*'
-
-    # If the above check fails, then the user is banned. Return false if they are permabanned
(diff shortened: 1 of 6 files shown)


app/helpers/user_api.rb
```diff
@@ -26,13 +26,18 @@ def self.get_user_info_pii(user_ids)
     HTTParty.post(competitor_info_path, headers: { WCA_API_HEADER => self.wca_token }, body: { ids: user_ids.to_a })
   end
 
-  def self.can_compete?(user_id)
+  def self.can_compete?(user_id, competition_start_date)
```
dunkOnIT, line 29: I think it's fine to pass in the comp start date here, but this did make me wonder if we should bring in the banned competitor logic to the registration_checker, which already has the competition info loaded in an instance variable.

FinnIckler: I think it's fine, the monolith also adds the competition_start_date in its function

FinnIckler (approved):

PR #591, opened 2024-05-27
Fix guests not being saved correctly

FinnIckler:
We moved guests out of the competing object, but forgot to change it here


FinnIckler:
@thewca-bot deploy staging

PR #600, opened 2024-06-10
Fix count being a string if it's 0

FinnIckler:


FinnIckler:
@thewca-bot deploy staging

FinnIckler:
Works

PR #602, opened 2024-06-10
Use shoryuken to allow workers to access Rails so we can trigger emails from the worker

FinnIckler:
This should be all the code changes that is needed, the rest should be config


FinnIckler:
@thewca-bot deploy staging

FinnIckler:
@thewca-bot deploy staging

FinnIckler:
@thewca-bot deploy staging

FinnIckler:
Successfully tested on staging

PR #607, opened 2024-06-20
Allow updating the waiting list through the update API and return history

FinnIckler:
two small changes needed for the admin PR


FinnIckler:
@thewca-bot deploy staging

PR #619, opened 2024-07-11
Add missing variables for worker

FinnIckler:
Now that the worker has full Rails access it needs these Environment variables. I added these for staging, but forgot to add them in production.


FinnIckler:
Ok, prod worker container is running now

PR #620, opened 2024-07-15
Add internal route to get registrations by attendee id

FinnIckler:


FinnIckler:
@thewca-bot deploy staging

PR #625, opened 2024-07-17
Increase Lambda Function Timeout to 10 seconds for first time starts

FinnIckler:
After I recently built the Lambda function again, our cold start increased over the 3 second threshold. So I'm increasing it again. Hot starts are still 100ms


FinnIckler:
Bonus: Bumps Ruby Version to 3.3

PR #626, opened 2024-07-17
Use find instead of find_by in internal route

FinnIckler:
These are not equivalent in dynamoid and using find_by requires an index to be there


FinnIckler:
@thewca-bot deploy staging

PR #628, opened 2024-07-26
Allow admins to update registrations

dunkOnIT:
Currently only competition organizers/delegates can update registrations - when we move from staging to prod, this will be a huge headache for us. This allows website admins to change update any registration in the new system.

Implementation detail: This switches to using the UserApi to check a user's permissions, rather than checking if a user is an org/delegate of a particular comp via the CompetitionApi - definitely the better way of doing it imo.

It does mean that a lot of added permissions mocks had to be added in tests. This is good, as a number of the existing permissions mocks were giving false-positives (mocking permissions for the user_id of the update request, not the submitted_by).

### DIFF
--- a/.rspec
+++ b/.rspec
@@ -1,1 +1,2 @@
+--format documentation
 --require spec_helper
--- a/app/services/registration_checker.rb
+++ b/app/services/registration_checker.rb
@@ -51,7 +51,8 @@ class RegistrationChecker
   class << self
     def user_can_create_registration!
       # Only an organizer or the user themselves can create a registration for the user
-      raise RegistrationError.new(:unauthorized, ErrorCodes::USER_INSUFFICIENT_PERMISSIONS) unless is_organizer_or_current_user?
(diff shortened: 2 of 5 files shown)


FinnIckler:
Mh.. the problem with this is, that we re-introduce the bug that an admin can update their own registration even if it isn't allowed by the organizers. I think we should also have to check for `allow_registration_edits? ` in the Competition API to mitigate this? Or should we just live with this 'bug'

dunkOnIT:
I think short-term we just live with this "bug" given that it's pretty low impact vs us not having full admin rights. Long-term we definitely want some kind of system that lets us enter "admin mode" consciously

dunkOnIT:
(Not sure if you were in the meeting, but Gregor was like "this is absolutely not negotiable, we have to have full admin access to everything in the website")

FinnIckler:
what's with all these commented out things in the tests?

dunkOnIT:
Ah good catch thanks! Those were commented out when I refactored the factories/setup in before blocks, forgot to remove them once the tests passed.

FinnIckler (approved):

PR #640, opened 2024-08-06
Add new Table for waiting List

FinnIckler:
This should be enough to get us started with porting over the tests


dunkOnIT:
For the `move_competitor` function - do we want to be passing in the target index, or target position? 

My suggestion is target position, because everywhere else I expect we will be using the competitor's position. Currently it is implemented as target index, however: 
```ruby
  def move_competitor(user_id, new_index)
```

dunkOnIT:
Will do! Busy with tests btw - should have them done by end of day my time (~4 hours or so]

dunkOnIT:
One complex case for waiting lists is the following: **User tries to accept multiple, sequential attendees off the waiting list in the same request**. [ie, selects waiting list position 1-4 and clicks "Accept", which fires off a bulk update request to our backend]

Currently, this fails because positions 2-4 are flagged as being accepting someone from the waiting list who is not the waiting list leader. I've considered a few different approaches for handling this: 
1. Remove the "must accept waiting list leader" check. _[At the very least, we should have a frontend warning for this - I have asked on Discord what user preference is]_
2. Keep track of the "new" waiting list leader in the backend. _[This is very messy/involves a lot of data sharing between functions at different levels, because the request does not include data about what the status _used_ to be - so we don't know from looking at an update whether it is being accepted from the waiting list or not]_
3. Add in the "old" registration data as part of the bulk update request, so that we can compare old/new competing statuses _[Seems bad unless we see a lot of other use cases for it]_
4. If, after validating a bulk update, all the error codes are -4011 [must accept waiting list leader], perform a check to see if all user_ids with 'accepted' statuses are match a sequential subset of the waiting list. This would also mean we have to prevent updating multiple lists at once in the frontend [ie, you can't accept from both the Pending and Waiting Lists at the same time]. _[My preferred approach]_

dunkOnIT:
Apart from the above query, tests for this PR are done as far as I'm concerned.

app/models/registration.rb
```diff
@@ -187,15 +171,18 @@ def update_payment_lane(id, iso_amount, currency_iso, status)
...
 
-  def update_waiting_list(lane, update_params)
-    update_params[:waiting_list_position]&.to_i
-    lane.add_to_waiting_list(competition_id) if update_params[:status] == 'waiting_list'
-    lane.accept_from_waiting_list if update_params[:status] == 'accepted'
-    lane.remove_from_waiting_list(competition_id) if update_params[:status] == 'cancelled' || update_params[:status] == 'pending'
-    lane.move_within_waiting_list(competition_id, update_params[:waiting_list_position].to_i) if
-      update_params[:waiting_list_position].present? && update_params[:waiting_list_position] != competing_waiting_list_position
-  end
+  # Dynamoid doesn't have a find_or_create_by so we need to use upsert
```
FinnIckler, line 174: oops this comment is no longer relevant, can you remove it?

app/controllers/registration_controller.rb
```diff
@@ -216,7 +216,8 @@ def validate_list_admin
 
   def list_admin
```
FinnIckler, line 217: Do we have tests that check if list_admin returns the correct data?

dunkOnIT:
Added tests for `list_admin`. A few notes here and in general:
- I've made the WaitingList functions generic [`add`, `remove`, etc], with the thought that as we add more lanes we can pass in the lane name as an argument to specify the lane, rather than defining a function for each lane (ie, `waiting_list.add(registration, :competing)`)
- Same goes for `Registration.waiting_list_position` - we can pass in the name of the lane to specify _which_ waiting list if needed
- I've modified the `list_admin` code to:
  - add `waiting_list_position` as a key in the `competing` hash by default (again, we can change this behaviour to be more specific when we add waiting lists for other lanes)
  - ONLY add `waiting_list_position` as a key on items which appear on the waiting list

app/controllers/registration_controller.rb
dunkOnIT: Great point! Tests added.

app/models/waiting_list.rb
```diff
@@ -0,0 +1,33 @@
...
+    end
+  end
+
+  def move_to_position(user_id, new_position)
+    raise ArgumentError.new('Target position out of waiting list range') if new_position > entries.length || new_position < 1
+
+    old_index = entries.find_index(user_id)
+    return if old_index == new_position-1
+
+    Rails.logger.debug(entries.inspect)
```
FinnIckler, line 29: logging

dunkOnIT: Oh yeah I think that was your line? Happy if we remove it

FinnIckler:
that should be it infrastructure wise, but I cannot test until the frontend PR is back, so I won't be able to do that today

FinnIckler:
@thewca-bot deploy staging

FinnIckler:
@thewca-bot deploy staging

FinnIckler:
@thewca-bot deploy staging

dunkOnIT (approved):

PR #652, opened 2024-08-23
Use authenticated route for querying competition API

dunkOnIT:
[Monolith 9809](https://github.com/thewca/worldcubeassociation.org/pull/9809) adds an internal/authenticated route where non-announced competitions can be queried via the API. This PR makes use of that route instead of the un-authenticated one


dunkOnIT:
@thewca-bot deploy staging

FinnIckler (approved):

PR #654, opened 2024-08-26
Raise errors on failed API calls

dunkOnIT:
Currently we haven't been checking for status code or raising errors when we don't get a 200 back from the monolith. This is a basic/non-comprehensive implementation of status codes for API endpoints that we call.

One thing I was wondering: If we get a 302/304 response, would that trigger a false-positive error message? Or is the server smart enough to still make it look like a 200?

### DIFF
--- a/app/controllers/registration_controller.rb
+++ b/app/controllers/registration_controller.rb
@@ -22,7 +22,7 @@ class RegistrationController < ApplicationController
   rescue Dynamoid::Errors::RecordNotFound
     render json: {}, status: :not_found
   rescue RegistrationError => e
-    render_error(e.http_status, e.error)
+    render_error(e.http_status, e.error, e.data)
   end
 
   def count
@@ -223,6 +223,8 @@ class RegistrationController < ApplicationController
     Metrics.registration_dynamodb_errors_counter.increment
     render json: { error: "Error getting registrations #{e}" },
            status: :internal_server_error
+  rescue RegistrationError => e
+    render_error(e.http_status, e.error, e.data)
   end
 
   def import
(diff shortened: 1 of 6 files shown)


FinnIckler:
I think httparty provides something like response.ok? To see if it is a 2xx or 3xx code

FinnIckler:
You can check the docs

dunkOnIT:
@thewca-bot deploy staging

lib/competition_api.rb
```diff
@@ -8,49 +8,29 @@ class CompetitionApi < WcaApi
...
-      when 404
-        Metrics.registration_competition_api_error_counter.increment
-        raise RegistrationError.new(404, ErrorCodes::COMPETITION_NOT_FOUND)
-      else
-        Metrics.registration_competition_api_error_counter.increment
-        raise RegistrationError.new(response.code.to_i, ErrorCodes::COMPETITION_API_5XX)
-      end
+    self.get_request("#{url(competition_id)}/qualifications")
+  rescue RegistrationError => e
+    if (e.data[:http_code] = 404)
```
gregorbg, line 20: I hope you're aware that in Ruby, comparisons are done using `==`?

In its current state, the `if` block reads "If you successfully *assigned* value 404 to the data, then proceed". In reality, that write operation will virtually never fail, and horrible things will happen in production.

gregorbg (changes requested):

lib/competition_api.rb
dunkOnIT: Lol yes I am well aware, copy-pasted typo. But it does illuminate the lack of tests on a 404 error, so I'll add those

PR #657, opened 2024-08-26
Prevent creation of duplicate registrations

dunkOnIT:
We weren't checking for duplicate registrations. This would likely still have yielded an error from DynamoDB, and wouldn't have been possible in the frontend, but obviously much better that it is explicitly handled by the `registration_checker`


dunkOnIT:
Must change this so that the worker rejects a request if the competing lane already exists

PR #663, opened 2024-08-28
Cache registrations_for_user

FinnIckler:
Cache registrations for each user and invalidate on create/update


FinnIckler:
fixes #662

FinnIckler:
@thewca-bot deploy staging

PR #670, opened 2024-09-02
Allow worker to connect to redis and fix it from dying if it can't

FinnIckler:
Reverts the hotfix, makes the worker more resilient and adds it to the security group that can access redis


dunkOnIT (approved):
Can't provide a truly informed review but seems reasonable based on the explanation in Slack

PR #673, opened 2024-09-04
Reimplment >= competitor limit check

dunkOnIT:
Previous version of this PR was reverted, as it triggered an error in live environments where Redis only returns strings


FinnIckler:
https://github.com/thewca/wca-registration/blob/3f7c7e8fc42491d6e45e0affef1cf591fae9a656/app/controllers/registration_controller.rb#L31
Here I just use .to_i not sure what is better

FinnIckler:
Anyway how you implement it, you should remove the to_i from my line

FinnIckler:
@dunkOnIT please fix the conflicts and remove the `to_i` where `Registration.accepted_competitors_count(competition_id)` is called elsewhere

FinnIckler (approved):

PR #674, opened 2024-09-05
Populate staging/dev registrations with a rake task

dunkOnIT:
This has been living in a side branch for a while, trying to bring it into main as it will make it much easier to have users test the waiting list.


dunkOnIT:
@thewca-bot deploy staging

PR #675, opened 2024-09-06
Add json response exceptions_app

FinnIckler:
So we can display a general error, currently we were returning HTML which crashed the error handling code


dunkOnIT:
@thewca-bot deploy staging

dunkOnIT:
@thewca-bot deploy staging

PR #684, opened 2024-09-11
Fix Build tag not being set correctly for deployments from prs

FinnIckler:
We fixed that in the monolith ages ago, but I didn't add it here...


gregorbg:
Yet another argument for merging both codebases ;)

PR #685, opened 2024-09-12
Bump Ruby Version to 3.3.5

FinnIckler:
to keep up with the monolith


FinnIckler:
merging as a dependency update

PR #686, opened 2024-09-13
Add rubocop-performance cop

FinnIckler:


FinnIckler:
merging as because code changes and tests pass

PR #694, opened 2024-09-20
Allow organizer to edit reg for a user with an accepted registration in another competition in the series

dunkOnIT:


FinnIckler:
I see you also added a fix for rejected not being seen as part of the already registered in series check. What about having a method in Registration called `attending?` or something that checks if the comepting status is not rejected or cancelled

dunkOnIT:
Have added a `might_attend?` property along the lines of what you were suggesting

dunkOnIT:
It does raise a question about whether or not we should let someone register for another comp in a series if they have any a `pending` or `waiting_list` status on an existing series comp registration

dunkOnIT:
@FinnIckler any chance you can take another look at this today? Would be good to get it deployed soon if it's ready to be

FinnIckler (approved):

PR #717, opened 2024-10-25
Add has_paid and payment_statuses in payload

FinnIckler:
For easier V2 -> V3 Migration


FinnIckler:
@thewca-bot deploy staging

PR #722, opened 2024-10-29
Allow organizers to approve registrations when there is no competitor limit

dunkOnIT:
This fixes an issue reported for TexasFMCChampionship2024

### DIFF
--- a/app/services/registration_checker.rb
+++ b/app/services/registration_checker.rb
@@ -174,9 +174,8 @@ class RegistrationChecker
       raise RegistrationError.new(:unprocessable_entity, ErrorCodes::INVALID_REQUEST_DATA) unless Registration::REGISTRATION_STATES.include?(new_status)
 
       competitor_limit = @competition_info.competitor_limit
-      accepted_competitors = Registration.accepted_competitors_count(@competition_info.competition_id)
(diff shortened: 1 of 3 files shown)


app/services/registration_checker.rb
```diff
@@ -173,7 +173,8 @@ def validate_update_status!
 
       raise RegistrationError.new(:unprocessable_entity, ErrorCodes::INVALID_REQUEST_DATA) unless Registration::REGISTRATION_STATES.include?(new_status)
       raise RegistrationError.new(:forbidden, ErrorCodes::COMPETITOR_LIMIT_REACHED) if
-        new_status == 'accepted' && Registration.accepted_competitors_count(@competition_info.competition_id) == @competition_info.competitor_limit
+        new_status == 'accepted' && Registration.accepted_competitors_count(@competition_info.competition_id) == @competition_info.competitor_limit &&
+        @competition_info.competitor_limit != 0
```
gregorbg, line 177: 1. `competitor_limit > 0`
2. You can put this check first, because the (relatively) expensive `Registration.accepted_competitors_count` doesn't even need to be run if there is no competitor limit

gregorbg (approved):

